PR #25, opened 2024-06-04
Allows users to manage multiple sites

vicr1:


kicr1 (approved):
Approved but with plan to investigate adding tests for the new GetSitesForUserFunction.cs and for the new logic added to SiteContextProvider.tsx.

PR #53, opened 2024-07-02
Initial frontend permissions implementation

joefarina:
Pertains to the following tickets, all of which are about reacting to roles in the UI (should be permissions):

APPT-48 - Check in role - disable check-in checkbox, show banner when without _booking:set-status_ permission

APPT-51 - Availability Manager role - hide 'templates/availability' navigation and disable routes when without _availability:get-setup_ permission

APPT-64 - Site Configuration Manager role - hide 'edit services' navigation and disable routes when without _site:get-config_ permission


src/client/src/Components/GuardedRouteHOC.tsx
vicr1, line 1: Would remove HOC from file name, I don't think this is immediately helpful and it is not consistent with how we have named other files

src/client/src/Components/UserNotification.tsx
```diff
@@ -0,0 +1,15 @@
+import "./UserNotification.css"
+
+export const UserNotification = ({title, handleClose}:{title:string, handleClose:()=>void}) => {
```
vicr1, line 3: Could we have a type for the props to keep it consistent

src/client/src/Types/Permissions.ts
```diff
@@ -0,0 +1,12 @@
+export enum Permissions {
+    CANGETSITES = "site:get-config",
```
vicr1, line 2: Block capitals is difficult for the human brain to decipher, I would remove the "can" from the start and make these Pascal case for readability

vicr1 (changes requested):

src/client/src/Views/DailyBookings.tsx
```diff
@@ -131,6 +141,11 @@ export const DailyBookings = ({ siteConfig, getBookings, setBookingStatus }: Dai
                     </div>
                 </div>
             </When>
+            <When condition={!hasPermission("booking:set-status") && showNotification}>
```
vicr1, line 144: Should use constant

vicr1 (changes requested):

vicr1 (approved):

PR #55, opened 2024-07-05
Migrate from create-react-app to nextjs

joefarina:
Initial switchover to nextjs, including testing framework. Old CRA App code is the same, it just runs inside a catch-all next.js App Route

https://nhsd-jira.digital.nhs.uk/browse/APPT-68


src/client/package.json
```diff
@@ -51,7 +49,12 @@
     "postcss": "^8.4.38"
```
kicr1, line 49: Do we need these overrides anymore? This was put in due to these being dependencies of CRA and CRA not updating them to later versions without the security vulnerabilities. Maybe with next.js these are managed or not used anymore. Probably worth a check.

kicr1 (approved):

kicr1 (approved):

PR #66, opened 2024-07-15
APPT-70 assign user roles

kicr1:


src/api/Nhs.Appointments.Persistance/UserStore.cs
```diff
@@ -19,4 +21,46 @@ public async Task<IEnumerable<RoleAssignment>> GetUserRoleAssignments(string use
         var userDocument = await cosmosStore.GetByIdAsync<UserDocument>(userId);
         return userDocument.RoleAssignments.Select(mapper.Map<RoleAssignment>);
     }
+    
+    public async Task SaveUserAsync(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
+    {
+        var documentType = cosmosStore.GetDocumentType();
```
vicr1, line 27: This can be moved as it is only used in one branch of the logic. Move it so it is close to the point it is used and in the most appropriate scope

vicr1 (approved):

PR #67, opened 2024-07-15
Appt 72 nextjs router

joefarina:
https://nhsd-jira.digital.nhs.uk/browse/APPT-72


src/client/src/Components/ServiceSelector.tsx
vicr1, line 1: Doesn't appear related to title of change

src/client/src/Components/UserNotification.tsx
vicr1, line 1: Unrelated change, make sure information is added to ticket otherwise these changes will not be tested

src/client/src/Views/TemplateListView.tsx
```diff
@@ -72,7 +72,7 @@ export const TemplateListView = () => {
                   <td>{t.name}</td>
                   <td>{getOpeningDayDetails(t)}</td>
                   <td>
-                    <Link to={`edit/${t.id}`}>
+                    <Link href={`templates/edit/${t.id}`}>
```
vicr1, line 75: Is this change of path intentional. Assume it is because previous routes worked with relative as opposed to absolute

src/client/src/index.css
vicr1, line 1: This change looks like it has the potential to affect styling across the entire site. That will increase the scope of testing for the ticket

src/client/src/pages/_app.tsx
vicr1, line 1: Is there a specific reason for having an underscore in the file name?

src/client/src/pages/site.tsx
vicr1, line 1: This seems somewhat confusing. The filename, component name and the underlying component being rendered don't quite agree on name. I was also wondering why we need these additional components (page components)

src/client/src/pages/templates/edit/[templateId].tsx
Comment (vicr1) [inline src/client/src/pages/templates/edit/[templateId].tsx:1]: These additional page component seem redundant

src/client/src/Components/ServiceSelector.tsx
joefarina: Next.js wont allow the importing of global css in components that isn't from node modules, it threw errors in the console so as part of this work I had to change from using global css to css modules for locally scoped component styles

src/client/src/pages/_app.tsx
joefarina: Yes, its next.js convention, it overrides their default App component: https://nextjs.org/docs/pages/building-your-application/routing/custom-app

src/client/src/pages/templates/edit/[templateId].tsx
Comment (joefarina) [reply src/client/src/pages/templates/edit/[templateId].tsx:1]: This file allows the dynamic routing to work (Pages Router only), as nextjs's router is file system based

src/client/src/index.css
joefarina: We were importing multiple levels of global css, the main index.css that only contained 2 CRA derived rules, then the App.css which contained the imported NHS frontend styles and our own overrides, then a couple of components with their own stylesheets. This was how CRA setup styles, now with nextjs its cleaner to have all the global css is in one place, NHS frontend styles are imported separately and explicitly and css modules handle locally scoped imports.

The order has changed slightly (the 2 former index.css rules now apply after the NHS frontend as opposed to before) and should produce no visual changes as long as index.css is imported _after_ the NHS frontend styles in _app.tsx.

Worth noting though that the original index.css styles do impact the font-family rule and aren't part of the NHS frontend, I'd say we remove them as they are likely just vestigial from CRA and removes the issue of global css coming from multiple files

src/client/src/Views/TemplateListView.tsx
joefarina: It is and your assumption is correct, although I anticipated nextjs to be able to handle relative paths, it doesnt work without "templates/". Adding a leading slash doesnt seem to work, this may be because both /edit and edit/[:id] route to the same component though other sources indicate relative paths arent  supported: https://stackoverflow.com/questions/69084689/relative-routing-in-next-js

vicr1 (approved):

PR #72, opened 2024-07-18
change to auth middleware to support different request types

vicr1:


kicr1 (commented):
Does the permissions snapshot test need to be updated now the RequiredPermissionAttribute takes a RequestInspector type?

vicr1:
> Does the permissions snapshot test need to be updated now the RequiredPermissionAttribute takes a RequestInspector type?

The addition of another property on to the attribute doesn't affect the test

kicr1 (approved):

PR #74, opened 2024-07-19
Appt 90,91,92 permission routing

joefarina:
Combines changes into one from 3 tickets, adds logout and email in header on site selection, guards previously allowed sub-routes, implements a catch-all for permission related errors to redirect the user

https://nhsd-jira.digital.nhs.uk/browse/APPT-90
https://nhsd-jira.digital.nhs.uk/browse/APPT-91
https://nhsd-jira.digital.nhs.uk/browse/APPT-92

### DIFF
--- a/src/client/src/Services/ApiClient.ts
+++ b/src/client/src/Services/ApiClient.ts
@@ -1,10 +1,12 @@
 import { useAuthContext } from "../ContextProviders/AuthContextProvider";
-import { getApiUrl } from "../configuration"
+import { getApiUrl } from "../configuration";
+import { useRouter } from "next/router";
 
 export const useAuthenticatedClient = () => {
     const { idToken, signOut } = useAuthContext();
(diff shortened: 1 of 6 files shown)


src/client/src/Services/ApiClient.ts
```diff
@@ -13,15 +15,21 @@ export const useAuthenticatedClient = () => {
           if(response.status === 401) {
               signOut();
           }
+          if(response.status === 403) {
+            router.push("/unauthorised");
```
vicr1, line 19: Rather than the apiClient being aware of the routing and mechanism the app is using for  unauthed requests it might be worth following the pattern of providing an unauthorised() callback that would have the advantage of separating the conerns and giving a single place to update if we change the route or mechanism for handling unauthorised requests

vicr1 (approved):

PR #75, opened 2024-07-22
Appt 96 create new user

kicr1:
- Add getRoles function to API to allow for client to get roles from cosmos
- Updated set user function validation
- Add edit user view which allows users to be able to add a new user. This view will also allow users to edit existing user (this will be done as part of a different ticket)
- Client and api unit tests
- Integration test for getRoles function


src/api/Nhs.Appointments.Api/Auth/SiteFromScopeInspector.cs
```diff
@@ -0,0 +1,32 @@
...
+        var body = await httpRequest.ReadBodyAsStringAndLeaveIntactAsync();
+
+        if (string.IsNullOrEmpty(body) == false)
+        {
+            try
+            {
+                using var jsonDocument = JsonDocument.Parse(body);
+                if (jsonDocument.RootElement.TryGetProperty("scope", out var scopeValue))
+                {
+                    return scopeValue.ToString().Replace("site:", "");
```
vicr1, line 21: Do we need to test to see if scopeValue starts with "site:"

src/client/src/Views/EditUser.tsx
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export const EditUserCtx = () => {
+    const { setUserRoles } = useUserService();
+    const { getRoles } = useRolesService();
+    const { site } = useSiteContext();
+    return ( <EditUser setUserRoles={setUserRoles} getRoles={getRoles} site={site!} />)
+}
+export const EditUser = ({setUserRoles, getRoles, site} : EditUserProps) => {
+    const router = useRouter();
```
vicr1, line 24: This could have been pushed into the ctx component and the redirect function could have been passed in via props, that would keep the component as reusable as possible and would have made it simpler to test (no need to mock the router module)

src/client/src/Views/EditUser.tsx
```diff
@@ -0,0 +1,163 @@
...
+    
+    const handleOnChange = (roleId: string) => {
+        selectedRoles?.includes(roleId) ?
+            setSelectedRoles( selectedRoles?.filter(r => r != roleId) ) :
+            setSelectedRoles( [...selectedRoles, roleId] );
+    }
+    
+    const validateFields = (user: string, selectedRoles: string[]) => {
+        const newValidationErrors: ValidationError[] = [];
+        const emailRegex = /[\w-.]+@([\w-]+\.)+[\w-]{2,4}/
```
vicr1, line 51: Could this regex be added as a consts file for use in other places?

tests/Nhs.Appointments.Api.Integration/Scenarios/RoleManagement/RoleManagementFeatureSteps.cs
```diff
@@ -0,0 +1,43 @@
...
+public sealed class RoleManagementFeatureSteps : BaseFeatureSteps
+{
+    private  HttpResponseMessage _response;
+    private HttpStatusCode _statusCode;
+    private GetRolesResponse _actualResponse;
+    
+    [Given("There are existing roles")]
+    public Task RoleAssignments()
+    {
+        return Task.CompletedTask;
```
vicr1, line 22: Maybe a comment here to explain why this is blank and that we're relying on basic steup for the tests to provide the roles

vicr1 (changes requested):

vicr1 (approved):

PR #79, opened 2024-07-24
Fixes SetBookingStatus siteId for permission checks

joefarina:
https://nhsd-jira.digital.nhs.uk/browse/APPT-129

### DIFF
--- a/src/api/Nhs.Appointments.Api/Functions/SetBookingStatusFunction.cs
+++ b/src/api/Nhs.Appointments.Api/Functions/SetBookingStatusFunction.cs
@@ -32,7 +32,7 @@ public class SetBookingStatusFunction : BaseApiFunction<SetBookingStatusRequest,
     [OpenApiSecurity("Api Key", SecuritySchemeType.ApiKey, Name = "Authorization", In = OpenApiSecurityLocationType.Header)]
     [OpenApiResponseWithBody(statusCode:HttpStatusCode.OK, "text/plain", typeof(SetBookingStatusResponse), Description = "Booking status successfully set")]
     [OpenApiResponseWithBody(statusCode:HttpStatusCode.NotFound, "text/json", typeof(ApiResult<object>), Description = "Booking not found")]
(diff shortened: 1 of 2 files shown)


kicr1 (approved):

src/api/Nhs.Appointments.Api/Functions/SetBookingStatusFunction.cs
```diff
@@ -32,7 +32,7 @@ public class SetBookingStatusFunction : BaseApiFunction<SetBookingStatusRequest,
     [OpenApiSecurity("Api Key", SecuritySchemeType.ApiKey, Name = "Authorization", In = OpenApiSecurityLocationType.Header)]
     [OpenApiResponseWithBody(statusCode:HttpStatusCode.OK, "text/plain", typeof(SetBookingStatusResponse), Description = "Booking status successfully set")]
     [OpenApiResponseWithBody(statusCode:HttpStatusCode.NotFound, "text/json", typeof(ApiResult<object>), Description = "Booking not found")]
-    [RequiresPermission("booking:set-status", typeof(NoSiteRequestInspector))]
+    [RequiresPermission("booking:set-status", typeof(SiteFromBodyInspector))]
```
vicr1, line 35: The body for this request does not contain site, so does this work?

vicr1 (changes requested):

src/api/Nhs.Appointments.Api/Functions/SetBookingStatusFunction.cs
joefarina: The request from the frontend includes the site in client/src/services/bookingservice.ts, which is then picked up by the SiteFromBodyInspector's JSON parsing, so its not in the SetBookingStatusRequest but is only used to determine authorisation

vicr1 (approved):
This works currently due to the fact that clients can optionally provide site id, which the client app does. We will review this later to validate if we're happy with that approach longer term

PR #86, opened 2024-08-01
Appt 141 - Auth and basic layout

vicr1:


src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Api.Functions
+{
+    public class GetUserProfileFunction(IValidator<EmptyRequest> validator, IUserContextProvider userContextProvider, ILogger<GetSitesForUserFunction> logger)
+    : BaseApiFunction<EmptyRequest, UserProfile>(validator, userContextProvider, logger)
+    {
+
+        [OpenApiOperation(operationId: "GetUserProfile", tags: new[] { "Utility" }, Summary = "Gets information about the signed in user")]
+        [OpenApiRequestBody("text/json", typeof(SetBookingStatusRequest))]
```
sambiramairelogic, line 23: Looks like this request type might be a copypasta mistake :)

src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
kicr1, line 1: If we are not using the GetSitesForUsers endpoint anymore would it be worth removing it now?

src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
vicr1: Indeed! I have resolved this in the very next PR which includes changes to this endpoint

src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
vicr1: I have removed it in the next PR when getting site becomes part of the user profile endpoint

PR #112, opened 2024-08-19
APPT-140: Create basic E2E tests

sambiramairelogic:
Installs playwright and writes some extremely simple log in / log out tests. 

This PR is just to nail down the config/setup, additional steps are coming in future PRs, such as: 
* Running these tests in the pipeline
* Deciding whether we want to use Page Object Models or not
* Getting Playwright to start up the Docker/.NET services for you if (and only if) they're not already running
* Testing on multiple browsers / viewports
* Tweaking config / abstraction patterns
* ADDING MORE MEANINGFUL TESTS :)


src/new-client/testing/log-in.spec.ts
```diff
@@ -0,0 +1,77 @@
+import { test, expect } from '@playwright/test';
+import env from './testEnvironment';
+
+const { TEST_USER_USERNAME, TEST_USER_PASSWORD } = env;
+
+test('Loads the site', async ({ page }) => {
```
kicr1, line 6: Could this be be asserted in the login test? i.e assert the page header before logging in.

src/new-client/testing/log-in.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+  await page.goto('/');
+
+  await expect(page).toHaveTitle('Appointment Management Service');
+});
+
+test('Has a log in link', async ({ page }) => {
```
kicr1, line 12: Could/should be covered by a component test? Get faster feedback on this by testing lower down.

src/new-client/testing/log-in.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    page.getByRole('heading', { name: 'You cannot access this site' }),
+  ).toBeVisible();
+  await expect(
+    page.getByText(
+      'You are currently not signed in. To use this site, please sign in.',
+    ),
+  ).toBeVisible();
+});
+
+test('Creates a token cookie upon sign in', async ({ browser }) => {
```
kicr1, line 59: This is testing implementation. Usually E2E tests are testing outcomes - i.e. ensure the user can login and is shown the correct page etc.

sambiramairelogic: Yeah, I almost didn't add it for that reason but I just wanted to prove that Playwright was capable of interacting with the browser storage. I was imagining we could write tests to remove the cookie then do a page transition or something. It was silly, I'll remove it.

src/new-client/testing/log-in.spec.ts
sambiramairelogic: I'm not sure I quite see what you mean here, sorry. Could you clarify a bit? 

(I am happy however to remove this test as it doesn't really add value)

src/new-client/testing/log-in.spec.ts
kicr1: Will this get implicitly tested by the login/logout tests in the sense that they will fail if that link is not visible? Or even add the assertion to that test to make it more explicit if needed? I'm just wondering on whether a separate test is needed for this?

src/new-client/testing/log-in.spec.ts
kicr1: In the sense of do we need a test to check that the site loads? Or could this assertion be part of the login test? i.e. before logging in check that the site has loaded. 

I guess I am coming form the mindset of E2E tests generally being more user journey specific so I would see this as the start of a user journey for example.

src/new-client/testing/log-in.spec.ts
sambiramairelogic: True enough - how about if I remove this test altogether, but add an extra assertion to the "logs in" test to prove that the "Log out" button is visible (& also maybe that "Log in" isn't)?

src/new-client/testing/log-in.spec.ts
sambiramairelogic: Yeah, I think I only added it from a "fail fast" mentality whilst working (i.e. if this test fails, we can stop right away because everything else is guaranteed to fail). 
How about I just remove it entirely?

sambiramairelogic: As discussed on our call, have removed :)

src/new-client/testing/log-in.spec.ts
sambiramairelogic: As discussed on our call, have removed but have added in assertions to the other tests

src/new-client/testing/log-in.spec.ts
sambiramairelogic: As discussed on our call, have removed :)

kicr1 (approved):

PR #115, opened 2024-08-20
APPT-170 Displays site list in ascending alphabetical order

kicr1:
- Displays site list in ascending alphabetical name order
- Unit test to check for correct order


src/new-client/src/app/lib/components/site-list.test.tsx
```diff
@@ -13,6 +13,23 @@ describe('<SiteList>', () => {
     expect(screen.queryByRole('link', { name: 'Site Beta' })).toBeVisible();
   });
 
+  it('renders sites in ascending alphabetical order', async () => {
+    const testSites: Site[] = [
```
sambiramairelogic, line 17: There's already a `mockSites` object we can import from `@testing/data` we could use instead?

src/new-client/src/app/lib/components/site-list.test.tsx
```diff
@@ -13,6 +13,23 @@ describe('<SiteList>', () => {
...
+      { id: '1001', name: 'Site Zulu', address: 'Zulu Street' },
+      { id: '1002', name: 'Site Lima', address: 'Lima Street' },
+      { id: '1003', name: 'Site November', address: 'November Street' },
+      { id: '1004', name: 'Site Beta', address: 'Beta Street' },
+    ];
+    render(<SiteList sites={testSites} />);
+    const list = screen.getByRole('list');
```
sambiramairelogic, line 24: This is entirely a nitpick so feel free to ignore me, but I think we could make this block slightly more readable using the `within` method from react testing library: 

`    
const siteNames = within(screen.getByRole('list')).getAllByRole('listitem');

expect(siteNames.length).toBe(4);
expect(siteNames[0]?.textContent).toEqual('Site Beta');
expect(siteNames[1]?.textContent).toEqual('Site Lima');
expect(siteNames[2]?.textContent).toEqual('Site November');
expect(siteNames[3]?.textContent).toEqual('Site Zulu');
`

sambiramairelogic: In fact that first check could even be `expect(siteNames).toHaveLength(4);` ..... sorry, obviously not an important change but just thought of it!

src/new-client/src/app/lib/components/site-list.test.tsx
kicr1: As discussed, I will just leave this as I have purposefully set up sites out of alpha order to check sorting which is different to the mocksite being used which will force changes to other tests.

src/new-client/src/app/lib/components/site-list.test.tsx
kicr1: All done :)

sambiramairelogic (approved):

PR #119, opened 2024-08-21
Consolidating page titles and breadcrumbs

vicr1:
Simple change to NHS Page and Error page to make it easier to ensure that breadcrumb end and page title match. Also fixes issue with trying to access a site that does not exist


src/new-client/src/app/lib/components/nhs-error-page.tsx
```diff
@@ -5,13 +5,21 @@ type Props = {
   title: string;
   message: string;
   breadcrumbs?: Breadcrumb[];
+  omitTitleFromBreadcrumb?: boolean;
```
sambiramairelogic, line 8: This is entirely a pedantic opinion so please feel free to ignore, but I found the naming of this slightly confusing - I originally thought it was removing the page title from a singular breadcrumb. Should it be plural - `omitTitleFromBreadcrumbs`? I also find logic in the negative case a bit harder to grapple with - could I suggest renaming this properly to follow the positive case instead, then we can also collapse some of the array pushing logic. Perhaps something along the lines of: 
image

sambiramairelogic (dismissed):
I've left one little comment with a suggested improvement, but am approving anyway so feel free to merge as-is if you disagree

src/new-client/src/app/lib/components/nhs-error-page.tsx
vicr1: Point taken on pluralization, and the collapsing of the logic, but I disagree on the naming of the prop. Given that the only time you would see it when it is not default I think that it is better that the positive should appear in that case as it easier for the reader of the markup to figure out what is going on
<NhsPage omitTitleFromBreadCrumbs={true} /> The mere presence of the prop tells the user what it is doing without them having to apply the inverse logic
<NhsPage addPageTitleBreadcrumb={false} /> asks me to apply the reverse logic to figure out what is happening.

sambiramairelogic: Yeah actually I see where you're coming from - I agree :)

We could optionally leave off the `={true}` as well as the markup doesn't need it. 
i..e `<NhsPage />` => it defaults to false
<`NhsPage omitTitleFromBreadcrumbs />` => sets it to true

There is actually an ES Lint rule that a lot of teams use which enforces this shorter syntax, but I'm not really fussed either way

sambiramairelogic (dismissed):

sambiramairelogic (approved):

PR #121, opened 2024-08-22
APPT-166: React Hook Form and NHS UK Frontend component pattern

sambiramairelogic:
* Creates a raft of nhs-uk styled form components.
* Establishes a pattern for using them with RHF by passing down the `ref` component using React's `forwardRef` feature. 
* Re-factors the Assign Roles Form and Find User Form to use this new pattern

### DIFF
--- a/src/new-client/src/app/site/[site]/users/manage/assign-roles-form.tsx
+++ b/src/new-client/src/app/site/[site]/users/manage/assign-roles-form.tsx
@@ -10,6 +10,7 @@ import {
   FormGroup,
   CheckBoxes,
   CheckBox,
+  ButtonGroup,
 } from '@nhsuk-frontend-components';
 
 type FormFields = {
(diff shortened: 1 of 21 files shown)


src/new-client/src/app/site/[site]/users/manage/assign-roles-form.tsx
```diff
@@ -40,64 +46,38 @@ const AssignRolesForm = ({
...
+      <div className="nhsuk-grid-row">
+        <div className="nhsuk-grid-row">
+          <div className="nhsuk-grid-column-one-half">
+            <Button type="submit">Confirm and save</Button>
+          </div>
+          <div className="nhsuk-grid-column-one-half">
+            <Button styleType="secondary" onClick={cancel}>
+              Cancel
+            </Button>
+          </div>
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/users/manage/assign-roles-form.tsx:80]: This styling isn't perfect - the two `one-half` columns cause more spacing between the buttons than I'd like. It's okay for now though, and I'd like to try to avoid writing custom inline styles if we can get away with it. Is it okay to leave this as it is for now and make a styling ticket for later?

vicr1 (approved):

PR #122, opened 2024-08-22
email notifications for user role change

jrowett:


src/api/Nhs.Appointments.Api/Consumers/UserRolesChangedConsumer.cs
```diff
@@ -0,0 +1,32 @@
...
+using Nhs.Appointments.Core.Messaging.Events;
+using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Api.Consumers
+{
+    public class UserRolesChangedConsumer : IConsumer<UserRolesChanged>
+    {
+        private readonly IUserRolesChangedNotifier _notifier;
+
+        public UserRolesChangedConsumer(IUserRolesChangedNotifier notifier)
```
vicr1, line 12: Use a primary constructor

src/api/Nhs.Appointments.Api/Functions/NotifyUserRolesChangedFunction.cs
```diff
@@ -0,0 +1,27 @@
...
+using Nhs.Appointments.Api.Consumers;
+
+namespace Nhs.Appointments.Api.Functions;
+
+public class NotifyUserRolesChangedFunction
+{
+    const string UserRolesChangedQueueName = "user-role-change";
+    readonly IMessageReceiver _receiver;
+
+    public NotifyUserRolesChangedFunction(IMessageReceiver receiver)
```
vicr1, line 15: Use a primary constructor (also removes the need for the private member variable)

src/api/Nhs.Appointments.Api/Functions/SetUserRolesFunction.cs
```diff
@@ -19,6 +19,11 @@ namespace Nhs.Appointments.Api.Functions;
 public class SetUserRolesFunction(IUserService userService, IValidator<SetUserRolesRequest> validator, IUserContextProvider userContextProvider, ILogger<SetUserRolesFunction> logger) 
     : BaseApiFunction<SetUserRolesRequest, EmptyResponse>(validator, userContextProvider, logger)
 {
+    protected readonly IUserService userService = userService;
```
vicr1, line 22: These seem completely superfluous

src/api/Nhs.Appointments.Api/Consumers/UserRolesChangedConsumer.cs
```diff
@@ -0,0 +1,32 @@
+﻿using MassTransit;
+using Nhs.Appointments.Api.Notifications;
+using Nhs.Appointments.Core.Messaging.Events;
+using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Api.Consumers
```
vicr1, line 6: We can use top level namespace declarations

src/api/Nhs.Appointments.Api/Notifications/ConsoleLogNotifications.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Nhs.Appointments.Core.Messaging;
+using System;
+using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Api.Notifications
```
vicr1, line 5: Can use top level namespace declaration

src/api/Nhs.Appointments.Api/Notifications/IUserRolesChangedNotifier.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Api.Notifications
```
vicr1, line 3: Top level namespace declaration

src/api/Nhs.Appointments.Api/Notifications/ServiceRegistration.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Nhs.Appointments.Api.Consumers;
+using Nhs.Appointments.Api.Functions;
+using Nhs.Appointments.Core.Messaging;
+using Nhs.Appointments.Core.Messaging.Events;
+using Notify.Client;
+using Notify.Interfaces;
+using System;
+
+namespace Nhs.Appointments.Api.Notifications
```
vicr1, line 11: Top level namespace declaration can be used

src/api/Nhs.Appointments.Api/Notifications/UserRolesChangedNotifier.cs
```diff
@@ -0,0 +1,39 @@
+﻿using MassTransit.Configuration;
+using Microsoft.Extensions.Options;
+using Notify.Client;
+using Notify.Interfaces;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Api.Notifications
```
vicr1, line 8: Top level namespace declaration can be used

src/api/Nhs.Appointments.Api/Notifications/UserRolesChangedNotifier.cs
```diff
@@ -0,0 +1,39 @@
+﻿using MassTransit.Configuration;
+using Microsoft.Extensions.Options;
+using Notify.Client;
```
vicr1, line 3: Unused using

src/api/Nhs.Appointments.Api/Notifications/UserRolesChangedNotifier.cs
```diff
@@ -0,0 +1,39 @@
+﻿using MassTransit.Configuration;
```
vicr1, line 1: Unused using

src/api/Nhs.Appointments.Api/Notifications/UserRolesChangedNotifier.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Nhs.Appointments.Api.Notifications
+{
+
+    public class UserRolesChangedNotifier : IUserRolesChangedNotifier
+    {
+        private readonly IAsyncNotificationClient _notificationClient;
+        private readonly string _emailTemplateId;
+
+        public UserRolesChangedNotifier(IAsyncNotificationClient notificationClient, IOptions<UserRolesChangedNotifier.Options> options)
```
vicr1, line 16: Primary constructor can be used

src/api/Nhs.Appointments.Api/local.settings.json
```diff
@@ -2,20 +2,24 @@
     "IsEncrypted": false,
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "AzureWebJobsServiceBus": "CONNECTION STRING TBC",
```
vicr1, line 5: This can be removed as we will not set this value in config

src/api/Nhs.Appointments.Api/local.settings.json
```diff
@@ -2,20 +2,24 @@
...
     "AuthProvider_Issuer": "http://localhost:8020",
     "AuthProvider_AuthorizePath": "connect/authorize",
     "AuthProvider_TokenPath": "connect/token",
     "AuthProvider_JwksPath": ".well-known/openid-configuration/jwks",
     "AuthProvider_ChallengePhrase": "ThisIsntRandomButItNeedsToBe43CharactersLong",
     "AuthProvider_ClientId": "nhs-appts-local",
-    "AuthProvider_ReturnUri": "http://localhost:7071/api/auth-return"
+    "AuthProvider_ReturnUri": "http://localhost:7071/api/auth-return",
+    "Notifications_Provider": "local",
+    "GovNotifyApiKey": "test",
```
vicr1, line 21: This can be removed

src/api/Nhs.Appointments.Core/IUserStore.cs
```diff
@@ -4,6 +4,7 @@ public interface IUserStore
 {
     Task<IEnumerable<RoleAssignment>> GetUserRoleAssignments(string userId);
     Task<string> GetApiUserSigningKey(string clientId);
+    Task<RoleAssignment[]> UpdateUserRoleAssignments(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments);
```
vicr1, line 7: Does this replace SaveUserAsync in which case can we remove it?

src/api/Nhs.Appointments.Core/Messaging/Events/UserRolesChanged.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Nhs.Appointments.Core.Messaging.Events
```
vicr1, line 1: Top level namespace can be used

src/api/Nhs.Appointments.Core/Messaging/IMessageBus.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Core.Messaging
```
vicr1, line 7: Top level namespace can be used

src/api/Nhs.Appointments.Core/Messaging/MassTransitBusWrapper.cs
```diff
@@ -0,0 +1,16 @@
+﻿using MassTransit;
+
+namespace Nhs.Appointments.Core.Messaging
```
vicr1, line 3: Top level namespace can be used

src/api/Nhs.Appointments.Core/Messaging/MassTransitBusWrapper.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class MassTransitBusWrapper : IMessageBus
+    {
+        private readonly IBus _bus;
+
+        public MassTransitBusWrapper(IBus bus)
```
vicr1, line 9: Primary constructor

src/api/Nhs.Appointments.Core/UserService.cs
```diff
@@ -1,14 +1,18 @@
...
+﻿using Nhs.Appointments.Core.Messaging;
+using Nhs.Appointments.Core.Messaging.Events;
+
+namespace Nhs.Appointments.Core;
 
 public interface IUserService
 {    
     Task<IEnumerable<RoleAssignment>> GetUserRoleAssignments(string userId);
     Task<string> GetApiUserSigningKey(string clientId);
+    Task UpdateUserRoleAssignmentsAsync(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments);
```
vicr1, line 10: Does this replace SaveUserAsync? If so we can remove it

src/api/Nhs.Appointments.Core/UserService.cs
```diff
@@ -19,10 +23,20 @@ public Task<string> GetApiUserSigningKey(string clientId)
...
-    public Task SaveUserAsync(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
+
+    public async Task UpdateUserRoleAssignmentsAsync(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
+    {
+        var oldRoles = await store.UpdateUserRoleAssignments(userId, scope, roleAssignments);
+
+        var rolesRemoved = oldRoles.Where(old => !roleAssignments.Any(r => r.Role == old.Role));
+        var rolesAdded = roleAssignments.Where(newRole => !oldRoles.Any(r => r.Role == newRole.Role));
+
+        await bus.Send(new UserRolesChanged { User = userId, Added = rolesAdded.Select(r => r.Role).ToArray(), Removed = rolesRemoved.Select(r => r.Role).ToArray()});
```
vicr1, line 34: This will send the role ids in the event, but at some point we will want resolve these to the display names for those roles before sending to gov notify so that the user doesn't receive a message containing role ids

src/api/Nhs.Appointments.Core/UserService.cs
```diff
@@ -19,10 +23,20 @@ public Task<string> GetApiUserSigningKey(string clientId)
...
+    }
+
+    public async Task SaveUserAsync(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
```
vicr1, line 37: Is this method now redundant? Also there is no need for this method to be marked as async  (which could be academic if it is being retired)

src/api/Nhs.Appointments.Persistance/UserStore.cs
```diff
@@ -21,7 +21,36 @@ public async Task<IEnumerable<RoleAssignment>> GetUserRoleAssignments(string use
...
+    /// <param name="scope"></param>
+    /// <param name="roleAssignments"></param>
+    /// <returns>The original role assignments prior to completion of the update operation</returns>
+    /// <exception cref="Exception"></exception>
+    public async Task<RoleAssignment[]> UpdateUserRoleAssignments(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
+    {
+        var originalDocument = await GetOrDefault(userId);
+        if (originalDocument == null)
+        {
+            throw new Exception($"Cannot update role assignments: user '{userId}' not found");
```
vicr1, line 38: This behaviour is different to the method it has replaced. This would cause an issue with the current functionality as it will now no longer be possible to assign roles to new users.

vicr1 (changes requested):

.gitignore
```diff
@@ -234,6 +234,7 @@ FakesAssemblies/
 *.crt
 /src/api/AppointmentsDataLoader
 /src/api/Nhs.Appointments.Api/jon.settings.json
+/src/api/Nhs.Appointments.Api/Properties/launchSettings.jon.json
```
kicr1, line 237: Is this supposed to be .jon.json?

.gitignore
```diff
@@ -234,6 +234,7 @@ FakesAssemblies/
 *.crt
 /src/api/AppointmentsDataLoader
 /src/api/Nhs.Appointments.Api/jon.settings.json
```
kicr1, line 236: Same as below - are these your specific settings? jon.settings.json

kicr1, line 236: src/api/Nhs.Appointments.Api/jon.settings.json
/src/api/Nhs.Appointments.Api/Properties/launchSettings.jon.json
Would these be better to be ignored in your git exclude file as they are you personal settings?

azure-pipelines.yml
```diff
@@ -225,7 +225,7 @@ stages:
                 $env:ARM_SUBSCRIPTION_ID="$(azure_subscription_id)"
                 cd $(Pipeline.Workspace)\drop_terraform\environments\dev
                 terraform init
-                terraform plan -no-color -input=false -var="HMAC_SIGNING_KEY=$(HMAC_SIGNING_KEY)" -var="API_KEYS=$(API_KEYS)" -var="APIM_SUBSCRIPTION_KEY=$(APIM_SUBSCRIPTION_KEY)"
+                terraform plan -no-color -input=false -var="HMAC_SIGNING_KEY=$(HMAC_SIGNING_KEY)" -var="API_KEYS=$(API_KEYS)" -var="APIM_SUBSCRIPTION_KEY=$(APIM_SUBSCRIPTION_KEY)" -var="gov_notify_api_key=$(gov_notify_api_key)" -var="user_roles_changed_email_template_id=$(user_roles_changed_email_template_id)"
```
kicr1, line 228: Is user_roles_changed_email_template_id a secret? If it is just a template id the it can be passed in as a normal TF_VAR. This pattern is only used for secret variables due to the way azdo secret pipeline variables work.

src/api/Nhs.Appointments.Core/Scope.cs
sambiramairelogic, line 1: I can see the utility of having a class to strongly type/name the properties of a scope, but I'm curious about why we're performing validation here and why we're returning null if a scope isn't valid. Is there a specific benefit to doing it this way rather than creating registering an IValidator with the Function to validate scopes on input? In general I usually prefer to fail fast and throw errors if things aren't valid rather than silently fail by returning null

src/api/Nhs.Appointments.Core/Messaging/Events/UserRolesChanged.cs
sambiramairelogic, line 1: This is a VERY minor pedantic point so please feel free to completely ignore, but in several places we're storing roles added/removed as string arrays. It's not immediately clear whether we're talking about the id or the label (until you read ahead to lines like `roles.Where(r => rolesAdded.Contains(r.Id)).Select(r => r.Name).ToArray();` 

I have a mild preference for making things more explicit, either by renaming things to `IdsAdded` / `IdsRemoved`, or like we've done with Scope create a Role type with `Id`/`Name` properties

src/api/Nhs.Appointments.Api/Notifications/UserRolesChangedNotifier.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// Depends on an email template hosted in Gov Notify. See template.txt for details 
+public class UserRolesChangedNotifier(IAsyncNotificationClient notificationClient, IRolesStore rolesStore, ISiteSearchService siteService, IOptions<UserRolesChangedNotifier.Options> options) : IUserRolesChangedNotifier
+{
+    public async Task Notify(string user, string siteId, string[] rolesAdded, string[] rolesRemoved)
+    {
+        var roles = await rolesStore.GetRoles();
+        var site = await siteService.GetSiteByIdAsync(siteId);
+        var siteName = site == null ? $"Unknown site ({siteId})" : site.Name;
+        var rolesAddedNames = roles.Where(r => rolesAdded.Contains(r.Id)).Select(r => r.Name).ToArray();
```
sambiramairelogic, line 18: Unless I'm just being blind and can't see it, I don't think we're validating the roles supplied by the function invoker before this point? 
This means if roles have been added that aren't present in the response from `await rolesStores.GetRoles();` we'll silently miss them. If we had a frontend bug which supplied the wrong ids for a role, we potentially wouldn't know about it. Perhaps we could throw in this case? Or add a validation layer earlier?

azure-pipelines.yml
sambiramairelogic: On the topic of template IDs, is it still the case that Gov.Notify offers us only a flat registry across all environments? Have we got a strategy for how we manage making updates to existing templates without affecting multiple environments?

vicr1: I believe the fact that we can specify the template id in the vars file, which can be environment specific should give us what we need

src/api/Nhs.Appointments.Api/Notifications/ServiceRegistration.cs
```diff
@@ -0,0 +1,54 @@
...
+        });
+
+        services.AddScoped<IAsyncNotificationClient>(x => new Notify.Client.NotificationClient(Environment.GetEnvironmentVariable("GovNotifyApiKey")));
+   
+        if (userNotificationsProvider == "local")
+        {
+            services
+                .AddTransient<IUserRolesChangedNotifier, UserRolesChangedNotifier>()
+                .AddScoped<IConsumer<UserRolesChanged>, UserRolesChangedConsumer>()
+                .AddScoped<IMessageBus, ConsoleLogWithMessageDelivery>();
```
vicr1, line 30: Does this mean that local running now relies on a connection to gov notify? Would it be more appropriate to have that as a separate config setting or is there a way in this configuration to have the actual sending via gov notify mocked out

src/api/Nhs.Appointments.Api/Notifications/UserRolesChangedNotifier.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    public async Task Notify(string user, string siteId, string[] rolesAdded, string[] rolesRemoved)
+    {
+        var roles = await rolesStore.GetRoles();
+        var rolesAddedNames = GetRoleNames(roles, rolesAdded);
+        var rolesRemovedNames = GetRoleNames(roles, rolesRemoved);
+
+        if (rolesAddedNames.Length == 0 && rolesRemovedNames.Length == 0)
+        {
+            // avoid sending pointless notifications when nothing has actually changed:
```
vicr1, line 21: Could we avoid the need for a comment by reversing the logic and nesting the main function logic, and having a private method HasRoleChanges so that the code reads appropriately without the need for a comment. I prefer to avoid comments 

e.g.
var hasRoleChanges = rolesAdded.Any() || rolesRemoved.Any()
if(hasRoleChanges) { ... do the sending }

vicr1 (changes requested):
Only thing outstanding is the issue with requiring gov notify when running locally. That should be optional and the local running version should work without requiring any connection to external services

azure-pipelines.yml
jrowett: correct, we can have multiple templates and can generate our own API keys with different permission sets

src/api/Nhs.Appointments.Core/Scope.cs
jrowett: this isn't a validator, it's just a string parsing utility method, put into a static class for readability

src/api/Nhs.Appointments.Core/Messaging/Events/UserRolesChanged.cs
jrowett: fixed

PR #128, opened 2024-08-28
APPT-100: User Management Permission Gating

sambiramairelogic:
Shows/hides UI elements that require either the `users:view` or `users:manage` permissions as appropriate. 
Ensures forbidden error messages are shown if routes are accessed without these permissions. 

In order to test this, we need a role which has `users:view` but does not have `users:manage`. Because we don't yet have this, I've created an "Auditer" role which is read only. I anticipate we might want to chat about this as a team and maybe make it a system role instead, but I can see it being useful. 

I've also created a raft of test users with the intention of only using them for the Playwright tests. They have the different permissions required in order to test all the user stories mentioned in the ticket (I need Manage&View, View, and Neither).

### DIFF
--- a/src/new-client/src/app/lib/components/nhs-page.test.tsx
+++ b/src/new-client/src/app/lib/components/nhs-page.test.tsx
@@ -8,6 +8,20 @@ const fetchUserProfileMock = fetchUserProfile as jest.Mock<
   Promise<UserProfile | undefined>
 >;
 
+jest.mock('@components/nhs-header-log-in', () => {
+  const MockNhsHeaderLogIn = () => {
+    return <button type="submit">log in</button>;
+  };
+  return MockNhsHeaderLogIn;
+});
+
+jest.mock('@components/nhs-header-log-out', () => {
+  const MockNhsHeaderLogOut = () => {
+    return <button type="submit">log out</button>;
+  };
+  return MockNhsHeaderLogOut;
+});
+
 describe('Nhs Page', () => {
   beforeEach(() => {
     fetchUserProfileMock.mockResolvedValue({


docker-compose.yml
sambiramairelogic, line 1: If we're happy to create multiple test users with different permissions for use in the Playwright tests, can we then move this JSON out of line and into a separate file?

mock-data/CosmosDbSeeder/items/index_data/global_roles.json
```diff
@@ -36,6 +36,19 @@
...
+      "description": "A user who can see everything (including admin-level data) but take no actions. This is a read-only role.",
+      "permissions": [
+        "site:get-config",
+        "site:get-meta-data",
+        "availability:get-setup",
+        "availability:query",
+        "booking:query",
+        "users:view"
+      ]
+    },
```
sambiramairelogic, line 51: As mentioned in the description, I anticipate a bit of discussion around this. Happy to make it a system role or whatever, I just need something with `users:view` but not `users:manage`.

src/new-client/src/app/site/[site]/users/manage/page.tsx
```diff
@@ -18,6 +18,11 @@ const AssignRolesPage = async ({ params, searchParams }: UserPageProps) => {
   const site = await fetchSite(params.site);
   const siteMoniker = site?.name ?? `Site ${params.site}`;
 
+  const permissions = await fetchPermissions(params.site);
+  if (!permissions.includes('users:manage')) {
+    throw new Error('Forbidden: You lack the necessary permissions');
+  }
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/users/manage/page.tsx:24]: We might be able to create some repeatable pattern for gating entire pages behind permissions - I think that might be a tad out of scope for this PR though?

src/new-client/testing/testEnvironment.ts
```diff
@@ -5,8 +5,12 @@ dotenv.config({ path: path.resolve(__dirname, 'playwright.env') });
...
 const environment = {
   BASE_URL: process.env.BASE_URL ?? '',
-  TEST_USER_USERNAME: process.env.TEST_USER_USERNAME ?? '',
-  TEST_USER_PASSWORD: process.env.TEST_USER_PASSWORD ?? '',
+  TEST_USER_1_USERNAME: process.env.TEST_USER_1_USERNAME ?? '',
+  TEST_USER_1_PASSWORD: process.env.TEST_USER_1_PASSWORD ?? '',
+  TEST_USER_2_USERNAME: process.env.TEST_USER_2_USERNAME ?? '',
+  TEST_USER_2_PASSWORD: process.env.TEST_USER_2_PASSWORD ?? '',
+  TEST_USER_3_USERNAME: process.env.TEST_USER_3_USERNAME ?? '',
+  TEST_USER_3_PASSWORD: process.env.TEST_USER_3_PASSWORD ?? '',
```
sambiramairelogic, line 13: I intended to wrap these up a bit to avoid so much noddy repetition - I intend to do it alongside introducing Page Object Model and other Playwright cleanups though

src/new-client/testing/user-management-permissions.spec.ts
sambiramairelogic, line 1: I'm aware there's lots of repetition in this suite - a lot of that can go once we introduce Page Object Model. 
I'm also aware there's an argument for saying this is too many e2e tests, but I think 3/5 are only testable here (the two that nav straight to the route rather than getting there through UI, and the one that views Church Lane then goes back to view Robin Lane instead). The other two are Happy Path user stories - so I think they're defensible. They don't currently seem to take long to run:
image

mock-data/CosmosDbSeeder/items/index_data/global_roles.json
kicr1: I think we have the flexibility to build up whatever roles we need. IMO I don't think for testing purposes we don't need to stick to the canned roles as defined in confluence. For the integration tests we created a role with the permissions need for running those so I don't see any issue with doing the same for the e2e testing pursposes.

src/new-client/src/app/site/[site]/users/users-page.tsx
```diff
@@ -5,34 +5,42 @@ import { Role, User } from '@types';
...
+  permissions: string[];
 };
 
-export const UsersPage = ({ users, roles }: Props) => {
+export const UsersPage = ({ users, roles, permissions }: Props) => {
   const isVisibleRole = (role: string) =>
     roles.find(r => r.id === role) !== undefined;
   const getRoleName = (role: string) =>
     roles.find(r => r.id === role)?.displayName;
+  const canSeeAdminControls = permissions.includes('users:manage');
```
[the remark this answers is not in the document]
Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/users/users-page.tsx:16]: Good spot - yeah might as well

PR #165, opened 2024-09-19
APPT-98: Remove User

sambiramairelogic:
Add a button to the User Management screens which allows the removal of all a user's roles from a specific site. 
If that is the only site for which they have roles, the entire user document is deleted. 
If they have roles at at least one other site, the document is instead patched so only the roles for the current site are removed. 

Please note I have not added any logic to prevent users from removing their own accounts, nor whether admins/"superusers" may remove other admins. We discussed this during refinement and I believe this needs further discussion and work as another ticket.


tests/Nhs.Appointments.Api.Integration/Scenarios/UserManagement/RemoveUserRoleAssignmentsSteps.cs
```diff
@@ -0,0 +1,78 @@
...
+        actualResult.Resource.RoleAssignments.Should().BeEquivalentTo(expectedRoleAssignments);
+    }
+
+    private async Task<UserDocument?> GetUserDocument(CosmosClient cosmosClient, string user)
+    {
+        var container = cosmosClient.GetContainer("appts", "index_data");
+        var document = container.GetItemLinqQueryable<UserDocument>(true)
+            .Where(d => d.Id == GetUserId(user))
+            .AsEnumerable()
+            .FirstOrDefault();
```
sambiramairelogic, line 74: This is a really weird to read a document, but `ReadItemAsync` will throw an exception if the document isn't found. This method is the best fit I could find for actually testing if something exists without excepting. Would welcome other suggestions if anyone knows a nicer way

kicr1: See if [ReadItemStreamAsync](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.container.readitemstreamasync?view=azure-dotnet) will work for what you need

This is in the docs for that method:
`The Stream operation only throws on client side exceptions. This is to increase performance and prevent the overhead of throwing exceptions. Check the HTTP status code on the response to check if the operation failed.`

One of the ReadItem methods would be preferable due to the RUs used for point reads vs queries although this is likely to be a lesser used function.

tests/Nhs.Appointments.Api.Integration/Scenarios/UserManagement/RemoveUserRoleAssignments.feature
```diff
@@ -0,0 +1,20 @@
+﻿Feature: Remove User roles from a Site
+
+  Scenario: Remove a user from a site
+    Given The following role assignments for 'test-one' exist
+      | Site | Roles                             |
+      | A    | canned:availability-manager       |
+    When I remove user 'test-one' from site 'A'
+    Then 'test-one' does not exist in the database
```
vicr1, line 8: Could we remove the implementation specific language (i.e. database) and perhaps change for then 'test-one' is no longer in the system

src/new-client/src/app/lib/components/remove-notification.tsx
```diff
@@ -0,0 +1,6 @@
+'use server';
+import { cookies } from 'next/headers';
+
+export const removeNotification = async () => {
+  cookies().delete('ams-notification');
```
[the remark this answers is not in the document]
vicr1: Moving to tech debt

vicr1 (approved):

PR #195, opened 2024-10-14
APPT-242: Add create availability card and tests

sambiramairelogic:
Creates a "Create Availability" card on the site landing page. Permission checking for this card is out of scope for this ticket, being covered instead in APPT-62.

### DIFF
--- a/src/new-client/src/app/site/[site]/site-page.test.tsx
+++ b/src/new-client/src/app/site/[site]/site-page.test.tsx
@@ -74,4 +74,19 @@ describe('Site Page', () => {
       screen.queryByRole('link', { name: 'Site Management' }),
     ).not.toBeInTheDocument();
   });
+
+  // TODO: Maybe parameterise these tests over permission/card pairs
+  // once we have more cards (i.e. during APPT-62 which adds in permission checking for this card)
+  it('shows the create availability page', () => {
+    const mockSite = mockSites[0];
+
+    render(<SitePage site={mockSite} permissions={mockAllPermissions} />);
+
+    expect(
+      screen.getByRole('link', { name: 'Create Availability' }),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByRole('link', { name: 'Create Availability' }),
+    ).toHaveAttribute('href', `${mockSite.id}/create-availability`);
+  });
 });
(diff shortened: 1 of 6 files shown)


src/new-client/src/app/site/[site]/site-page.test.tsx
```diff
@@ -74,4 +74,19 @@ describe('Site Page', () => {
       screen.queryByRole('link', { name: 'Site Management' }),
     ).not.toBeInTheDocument();
   });
+
+  // TODO: Maybe parameterise these tests over permission/card pairs
```
Comment (kicr1) [inline src/new-client/src/app/site/[site]/site-page.test.tsx:78]: Would it be worth doing this now as I assume we will be getting more cards? If we aren't having any more cards will we ever come to do this and therefore do we need the comment?

src/new-client/src/app/site/[site]/site-page.tsx
```diff
@@ -23,28 +23,33 @@ export const SitePage = ({ site, permissions }: SitePageProps) => {
...
-                title="Site Management"
-                description="Assign accessibility attributes to this site"
-              />
-            </li>
-          )}
-        </ul>
-      )}
+      <ul className="nhsuk-grid-row nhsuk-card-group">
+        <li className="nhsuk-grid-column-one-half nhsuk-card-group__item">
+          <Card
```
Comment (kicr1) [inline src/new-client/src/app/site/[site]/site-page.tsx:28]: Do we need to add a permissions to this card? I believe there are permissions for this or is this currently under review and therefore we need to wait?

kicr1 (approved):

src/new-client/src/app/site/[site]/site-page.test.tsx
Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/site-page.test.tsx:78]: I did consider it, but the other two cards are currently permission-gated but this one isn't. 
I'm expecting the parameterised tests to involve the permissions (i.e. something like it.each("appears when user has permission"(card, permission) and another "it.each("doesn't appear when user doesn't")(card) etc. ) so I'd like to do the parameterisation once all 3 cards have permission gating.

src/new-client/src/app/site/[site]/site-page.tsx
Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/site-page.tsx:28]: Yeah, the [ticket](https://nhsd-jira.digital.nhs.uk/browse/APPT-242) explicitly mentions that permissions should be done in a separate, linked [ticket ](https://nhsd-jira.digital.nhs.uk/browse/APPT-62). I'm not sure why we didn't prep these tickets together, but for whatever reason we haven't

kicr1 (approved):

PR #223, opened 2024-10-24
allow provisional bookings to be made without providing contact details

jrowett:


src/api/Nhs.Appointments.Core/BookingsService.cs
```diff
@@ -69,15 +71,21 @@ public Task<IEnumerable<Booking>> GetBookingByNhsNumber(string nhsNumber)
 
             if (canBook)
             {
+                booking.Created = _time.GetLocalNow().DateTime;
```
vicr1, line 74: I think this should be UTC as it is server side data that we use in processing

src/api/Nhs.Appointments.Persistance/BookingCosmosDocumentStore.cs
```diff
@@ -10,21 +10,23 @@ public class BookingCosmosDocumentStore : IBookingsDocumentStore
...
     public Task<IEnumerable<Booking>> GetInDateRangeAsync(DateTime from, DateTime to, string site)
     {
         return _bookingStore.RunQueryAsync<Booking>(b => b.Site == site && b.From >= from && b.From <= to);
     }
 
-    public async Task<IEnumerable<Booking>> GetCrossSiteAsync(DateTime from, DateTime to)
+    public async Task<IEnumerable<Booking>> GetCrossSiteAsync(DateTime from, DateTime to, bool provisional = false)
     {
-        var bookingIndexDocuments = await _indexStore.RunQueryAsync<BookingIndexDocument>(i => i.From >= from && i.From <= to);
+        var bookingIndexDocuments = await _indexStore.RunQueryAsync<BookingIndexDocument>(i => i.From >= from && i.From <= to && i.Provisional == provisional);
```
vicr1, line 29: We might want to look at this query and decide if it is better to do this on cosmos or in code. If we only expect a small number of provisionals to come in we might actual find that the RUs are lower if we don't have this on the query. Worth checking by trying it in the cosmos emu

src/api/Nhs.Appointments.Persistance/BookingCosmosDocumentStore.cs
```diff
@@ -90,6 +92,23 @@ public async Task<bool> UpdateStatus(string bookingReference, string status)
...
+        {
+            return false;
+        }
+
+        var updateStatusPatch = PatchOperation.Replace("/provisional", false);
+        await _bookingStore.PatchDocument(bookingIndexDocument.Site, bookingReference, updateStatusPatch);
+        await _indexStore.PatchDocument("booking_index", bookingReference, updateStatusPatch);
+
+        var addContactDetailsPath = PatchOperation.Add("/contactDetails", contactDetails);
+        await _bookingStore.PatchDocument(bookingIndexDocument.Site, bookingReference, addContactDetailsPath);
```
vicr1, line 108: Patch document can accept multiple Patch operation and will apply them atomically, so we can do this for the bookingStore update

vicr1 (changes requested):

vicr1 (approved):

PR #230, opened 2024-10-28
Appt 327 - Add information for citizens text area

ste-banks:
New free text area for adding site information that would be useful for citizens. The GetSite endpoint and (POST) SiteAttributes endpoint have been updated to include a _scope_ parameter so to only update / retrieve specific attributes for that site. A default value of "*" has been set for the scope parameter which will return all attributes for that site if a specific scope hasn't be requested.


src/new-client/src/app/lib/services/appointmentsService.ts
```diff
@@ -181,7 +184,7 @@ export const saveUserRoleAssignments = async (
 
 export const saveSiteAttributeValues = async (
   site: string,
-  attributeValues: AttributeValue[],
+  attributeValues: SetAttributes,
```
sambiramairelogic, line 187: Very minor thing, but just for consistency, elsewhere we've explicitly named request/response types (i.e. `ApplyAvailabilityTemplateRequest`. Could we make this `SetAttributesRequest` or something similar?

src/new-client/src/app/site/[site]/details/edit-information-for-citizens/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export type PageProps = {
+  params: {
+    site: string;
+  };
+};
+
+const Page = async ({ params }: PageProps) => {
+  const site = await fetchSite(params.site);
+  const siteMoniker = site?.name ?? `Site ${params.site}`;
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/edit-information-for-citizens/page.tsx:13]: You can ditch the siteMoniker stuff now, since the result of `fetchSite` can be relied upon

src/new-client/src/app/site/[site]/details/site-details-page.tsx
```diff
@@ -16,7 +16,13 @@ const SiteDetailsPage = async ({ site, permissions }: Props) => {
   const accessibilityAttributeDefinitions = attributeDefinitions.filter(ad =>
     ad.id.startsWith('accessibility'),
   );
-  const siteAttributeValues = await fetchSiteAttributeValues(site.id);
+  const siteDetails = await fetchSite(site.id);
+  const informationForCitizenAttribute =
+    siteDetails === undefined
+      ? ''
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/site-details-page.tsx:22]: As with siteMoniker, you should now be able to ditch the undefined check here I think?

src/new-client/src/app/site/[site]/details/site-details-page.tsx
```diff
@@ -54,6 +61,32 @@ const SiteDetailsPage = async ({ site, permissions }: Props) => {
           </Button>
         </Link>
       )}
+
+      <h3>Information for citizens</h3>
+      <p>Instructions to be show to people when they arrive</p>
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/site-details-page.tsx:66]: Minor typo here? show / shown

src/new-client/src/app/lib/services/appointmentsService.ts
```diff
@@ -235,3 +238,25 @@ export const saveAvailability = async (
...
+) => {
+  const response = await appointmentsApi.post(
+    `sites/${site}/attributes`,
+    JSON.stringify(attributeValues),
+  );
+
+  // TODO: Notification?
+  handleBodyResponse(response);
+  revalidatePath(`/site/${site}/information-for-citizens`);
```
sambiramairelogic, line 261: I might've misread, but I don't think this is the right path? Looks like it should be `/site/${site}/details`? Admittedly, I'm not 100% certain these calls are actually having the intended effect yet so I do intend to investigate all our caching stuff at some point

src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    await setSiteInformationForCitizen(site, payload);
+
+    replace(`/site/${site}/details`);
+  };
+
+  return (
+    <form onSubmit={handleSubmit(submitForm)}>
+      <FormGroup legend="Information for citizen">
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:53]: Should this be plural citizens?

src/new-client/src/app/site/[site]/details/edit-information-for-citizens/edit-information-for-citizens-page.tsx
```diff
@@ -0,0 +1,32 @@
...
+    site,
+    'site_details',
+  );
+  const siteInformation = informationForCitizens.find(
+    i => i.id === 'site_details/info_for_citizen',
+  );
+
+  return (
+    <>
+      <div className="nhsuk-form-group">
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/edit-information-for-citizens/edit-information-for-citizens-page.tsx:20]: There is a `<FormGroup/>` component that accepts a legend and a validation error. It could be expanded to accept a hint too if you wanted to use it here but I'm happy either way :)

src/new-client/src/app/lib/services/appointmentsService.ts
sambiramairelogic, line 1: Looks like this file is now reporting a linting error now `AttributeValue` is no longer used - our pre-commit hook should have caught this during the linting step so this might be a signal that it's not working properly. I'll look into it some time...

src/new-client/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@playwright/test": "^1.48.2",
+    "@testing-library/dom": "^10.4.0",
```
sambiramairelogic, line 33: Forgive me if I'm mistaken, but I don't think we need to add this package?

src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <div className="nhsuk-form-group">
+          <TextArea
+            label="What information would you like to include?"
+            {...register('informationForCitizen')}
```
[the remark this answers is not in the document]
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:57]: Good point on the validation. The back end will return a 400 if it's empty but there could be a valid use case where a site would want to remove this information meaning we'd need to allow for an empty string. E.g. if it's outdated. Not sure if it's worth updating the back end to remove the attribute if the user passes up an empty string?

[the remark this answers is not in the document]
Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:57]: Mhm yeah - a REST purest might argue that we should create a DELETE endpoint for this case but I'd be happy with just updating the existing endpoint to accept the empty string. Might be worth a quick team bunker at some point just to check we all agree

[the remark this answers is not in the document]
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:57]: Will resolve this as per the chat in slack... allow empty strings and remove in the document if it is empty.

sambiramairelogic (approved):

PR #235, opened 2024-10-30
Appt 357 - Get site meta data

ste-banks:
Implementing the endpoint to get a sites _information for citizen attribute value_ (if applicable).

I've had to update the route from _/sites/meta?site={siteId}_ to _/sites/{siteId}/meta_ because the former (for reasons unknown) kept hitting the GetSite endpoint for me but when I updated it to the latter it worked just fine.


tests/Nhs.Appointments.Api.Integration/Scenarios/SiteManagement/GetSiteMetaData.feature
```diff
@@ -0,0 +1,19 @@
...
+    Then the correct site meta data is returned
+      | Site   | AdditionalInformation |
+      | Site-A | test information      |
+
+  Scenario: Retrieve site meta data when it hasn't been set
+    Given The following sites exist in the system
+      | Site | Name   | Address      | Attributes             | Longitude | Latitude |
+      | A    | Site-A | 1A Site Lane | def_one/attr_one=false | -60       | -60      |
+    When I request site meta data for site 'A'
+    Then the correct site meta data is returned
```
vicr1, line 17: It would be nice to have something that reads more naturally here, so I would prefer a specific step instead of reusing the existing one . e.g.
Then no site meta is returned

tests/Nhs.Appointments.Api.Integration/Scenarios/SiteManagement/GetSiteMetaDataFeatureSteps.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            var siteId = GetSiteId(site);
+            Response = await Http.GetAsync($"http://localhost:7071/api/sites/{siteId}/meta");
+        }
+
+        [Then("the correct site meta data is returned")]
+        public async Task Assert(DataTable dataTable)
+        {
+            var row = dataTable.Rows.ElementAt(1);
+            var info = row.Cells.ElementAt(1).Value == "__empty__"
```
vicr1, line 26: See previous comment regarding making a step that reads more natually

vicr1 (changes requested):

vicr1 (approved):

PR #241, opened 2024-10-31
APPT-XXXX: Loading spinners

sambiramairelogic:
Creates Loading Spinner components according to [the nhsuk-frontend backlog](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/255) and uses them to replace submit buttons during form submissions.


ste-banks (approved):

kicr1 (commented):
Possibly a stupid question but is there anyway to test the spinner? May not be needed.

sambiramairelogic:
> Possibly a stupid question but is there anyway to test the spinner? May not be needed.

It's a good question, certainly doable. 
I think component tests would be fairly pointless as there's no behaviour to test, and testing for css is icky. 
We could test that the spinners are used by the form components by mocking out the `useForm<>` object and setting `isSubmitting` and `isSubmitSuccessful` to true.

There might be a smart way to handle the async nature of all this. Naturally we don't know how long the submission is going to take, and our tests await each assertion to accommodate this delay. So writing "First you see the see then you see the next page" might be flaky. I'll have a quick play with mocking the useForm return and see how it goes :)

kicr1 (approved):

PR #249, opened 2024-11-04
Appt 270 additional site info validation

ste-banks:
- Adding a max length of 150 to the site information text area
- Including remaining character count beneath the text area
- Adding checks to see if the text contains certain special characters or a URL and if so, fail validation, don't allow for submission and display a validation message

### DIFF
--- a/src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
+++ b/src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
@@ -22,11 +22,12 @@ const AddInformationForCitizensForm = ({
   site: string;
 }) => {
   const { replace } = useRouter();
-  const { register, handleSubmit } = useForm<FormFields>({
+  const { register, handleSubmit, formState } = useForm<FormFields>({
     defaultValues: {
       informationForCitizen: information,
     },
   });
+  const { errors } = formState;
 
   const cancel = () => {
     replace(`/site/${site}/details`);
@@ -48,13 +49,35 @@ const AddInformationForCitizensForm = ({
     replace(`/site/${site}/details`);
   };
 
+  const isValidTextInput = (text: string): boolean => {
+    const urlRegex = new RegExp(
+      '([a-zA-Z0-9]+://)?([a-zA-Z0-9_]+:[a-zA-Z0-9_]+@)?([a-zA-Z0-9.-]+\\.[A-Za-z]{2,4})(:[0-9]+)?([^ ])+',
+    );
+    if (urlRegex.test(text)) {
+      return false;
+    }
+
+    const specialCharacterRegex = /^[-\w \.\,\-]+$/;
+    return specialCharacterRegex.test(text);
+  };
+
   return (
     <form onSubmit={handleSubmit(submitForm)}>
-      <FormGroup legend="Information for citizens">
+      <FormGroup
+        legend="Information for citizens"
+        error={errors.informationForCitizen?.message}
+      >
         <div className="nhsuk-form-group">
           <TextArea
             label="What information would you like to include?"
-            {...register('informationForCitizen')}
+            maxLength={150}
+            {...register('informationForCitizen', {
+              validate: value => {
+                if (!isValidTextInput(value)) {
+                  return "Text cannot contain a URL or special characters outside of '.' ',' '-'";
+                }
+              },
+            })}
           />
         </div>
       </FormGroup>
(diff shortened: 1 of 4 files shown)


src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
```diff
@@ -48,16 +52,57 @@ const AddInformationForCitizensForm = ({
...
+        <FormGroup
+          legend="Information for citizens"
+          error={errors.informationForCitizen?.message}
+        >
+          <div className="nhsuk-form-group">
+            <TextArea
+              label="What information would you like to include?"
+              maxLength={maxLength}
+              {...register('informationForCitizen', {
+                validate: value => {
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:87]: There's a way we can handle this within the React Hook Form framework that avoids needing write it yourself, or introduce the extra state variables to track input length. 

The [register API](https://www.react-hook-form.com/api/useform/register/) exposes `maxLength` and `pattern` methods which will achieve the same result, and this will automatically be called when the form is submitted. 

Something along the lines of: 
`{...register("informationForCitizens", { maxLength: 150, pattern: urlRegex })}` should suffice

You can also do `.....maxLength: { value: 150, message: "custom message for this rule" }, pattern: {value: urlRegex, message: 'custom-other-message"}}` to customise it a bit more.

src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
```diff
@@ -48,16 +52,57 @@ const AddInformationForCitizensForm = ({
...
+                  handleTextInputUpdate(e.target.value.length);
+                },
+              })}
+            />
+          </div>
+          <div
+            className="nhsuk-hint nhsuk-character-count__message"
+            id="more-detail-info"
+          >
+            You have {maxLength - textInputLength} characters remaining
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:102]: We can avoid the extra state variable by using [watch](https://www.react-hook-form.com/api/useform/watch/) exposed by RHF. 
something like 
`const infoWatch = watch('informationForCitizens');`
.... `you have {maxLegth - infoWatch.length} ...`

src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
```diff
@@ -48,16 +52,57 @@ const AddInformationForCitizensForm = ({
...
+            <TextArea
+              label="What information would you like to include?"
+              maxLength={maxLength}
+              {...register('informationForCitizen', {
+                validate: value => {
+                  if (!isValidTextInput(value)) {
+                    return "Text cannot contain a URL or special characters outside of '.' ',' '-'";
+                  }
+                },
+                onChange: e => {
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:92]: If we use the RHF methods I've mentioned elsewhere, we shouldn't need this onChange method at all :)

sambiramairelogic (changes requested):
(as per comments)

src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx
```diff
@@ -48,16 +52,57 @@ const AddInformationForCitizensForm = ({
     replace(`/site/${site}/details`);
   };
 
+  const isValidTextInput = (text: string): boolean => {
+    const urlRegex = new RegExp(
+      '([a-zA-Z0-9]+://)?([a-zA-Z0-9_]+:[a-zA-Z0-9_]+@)?([a-zA-Z0-9.-]+\\.[A-Za-z]{2,4})(:[0-9]+)?([^ ])+',
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:57]: This might be a good opportunity to introduce [schema driven validation](https://www.react-hook-form.com/get-started/#SchemaValidation) so we could probably do something like `yup().string().not().email()` and save having to write a regex ourselves. 
Admittedly it's scope creep, it's just something I've been meaning to get around to doing at some point so thought I'd mention it :)

Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:57]: Ooh actually I've just remember we have a `constants.tsx` file that already has an `EMAIL_REGEX` in it. On the one hand I'd rather scrap this file and replace it with schema validation at some point, but on the other hand if we want to hold off for now and introduce schemas at a later date we should probably avoid the repetition and use this here. We could add a second variable like `SPECIAL_CHARACTERS_REGEX` or something to handle the other cases, whatever suits :)

Comment (ste-banks) [reply src/new-client/src/app/site/[site]/details/edit-information-for-citizens/add-information-for-citizens-form.tsx:57]: Didn't realise that file existed, I'll stick these two in there and we could always add the schema driven validation at a later date. It's also a URL regex not another email one so there will be 3 in there now :)

PR #250, opened 2024-11-04
APPT-232: Improve wizard validation

sambiramairelogic:
This PR cleans up most of the immediately obvious validation errors in the wizard. Chiefly: 

- Errors are no longer set manually through the `setError method`, so they no longer persist between step transitions
- Trying to click Continue without any checkboxes selected on the Days/Services steps no longer required 2 clicks to "fix" the error message
- Since the "Days of the Week" step should be skipped when Session Type is Single, clicking "Back" on the step after it no longer erroneously shows you the step (this has had to mean dragging the Back Link code back into each step and out of the generic <WizardStep>.)
- Services in the summary step are now displayed using their label rather than their value
- The summary step no longer shows a row for "Days of the Week" when Session Type is single
- The capacity calculation no longer says "No Capacity" if any of its input fields are dirty. This was caused by the inputs changing `01` into `"01"`, which then failed the `Number.isInteger()` checks. 

Other changes: 

- `StartDate`, `EndDate`, `Session.StartTime`, and `Session.EndTime` are no longer validated by each individual component. They are now wrapped in a [Controller](https://www.react-hook-form.com/api/usecontroller/controller/) so they can be treated as a single field.


ste-banks:
Do we need to add any extra validation tests for the time-and-capacity-step component? Noticed there are validation checks on start time being before end time but no corresponding tests for them.

sambiramairelogic:
> Do we need to add any extra validation tests for the time-and-capacity-step component? Noticed there are validation checks on start time being before end time but no corresponding tests for them.

Yeah that's a fair shout, I'll add some in

ste-banks (approved):

PR #258, opened 2024-11-06
Appt-390 - Updating special character regex to not allow underscores

ste-banks:


sambiramairelogic (approved):

sambiramairelogic:
I don't mind this going in without, but it might be worth considering expanding this test to be an `it.each()` to get a regression of what this regex should accept 
image

ste-banks:
> I don't mind this going in without, but it might be worth considering expanding this test to be an `it.each()` to get a regression of what this regex should accept image

Good shout, have updated it to contain 3 separate tests using it.each

sambiramairelogic (approved):

PR #273, opened 2024-11-13
APPT-229 - Preventing users from modifying their own role assignments when logged in

ste-banks:


src/new-client/src/app/site/[site]/users/users-page.tsx
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/users/users-page.tsx:1]: Just removing the button isn't enough, we need to guard against the user manually entering the route in the browser. On `users/remove/page.tsx` we're doing this like so:
image

Comment (ste-banks) [reply src/new-client/src/app/site/[site]/users/users-page.tsx:1]: Ah yeah good point, I'll update that

sambiramairelogic (dismissed):

sambiramairelogic (approved):

PR #284, opened 2024-11-19
reschedule notifications

jrowett:


mock-data/CosmosDbSeeder/items/index_data/notification_config.json
```diff
@@ -61,6 +61,30 @@
       "emailTemplate": "2bb90de1-2d7d-45b9-bc6e-88d0c04f0021",
       "smsTemplate": "2ef9995c-32b9-4088-a630-97c2546062d0",
       "eventType": "BookingCancelled"
+    },
+    {
+      "services": [ "COVID:12_15", "COVID:16_17", "COVID:18_74", "COVID:5_11", "COVID:75" ],
+      "emailTemplate": "",
```
vicr1, line 67: Will this cause issues, i.e. will this fail in our deployed environments if template ids are not provided?

jrowett: it won't result in a 500, the notification will go to the queue and the queue handler will fail. so you'll see some deadletters but API clients will be fine. he said confidently

jrowett: if you'd rather not see those errors in the Azure logs we can set up placeholder templates then replace them when those tickets get picked up

jrowett: placeholders added

vicr1 (approved):

PR #287, opened 2024-11-20
APPT-368: Backend for recording "Availability Created Events"

sambiramairelogic:
To populate the "Create Availability" page where the users can see previous submissions of the Create Availability Wizard, we decided to treat previous submissions like an audit trail. 

This PR creates the concept of an "Availability Created Event", which captures submissions of the wizard either for Single Sessions OR for Templates. I have decided to store one document per submission (rather than build up a massive document per site), but to store both Single Session events and Template events in a common DocumentType for the purposes of querying speed. 

There is also a new API to GET all the AvailabilityCreatedEvents for a given site. Even though this is a read action, I've given it the availability:set permission because it's intrinsically linked to the availability creation actions. 

There will be a follow-up PR with the frontend changes this ticket.


tests/Nhs.Appointments.Api.Integration/Scenarios/CreateAvailability/ApplyAvailabilityTemplate.feature
```diff
@@ -1,23 +1,26 @@
...
-          | Tomorrow_+2 | 09:00 | 10:00 | COVID    | 5           | 1        |
+  Scenario: Can apply an availability template
+    Given there is no existing availability
+    When I apply the following availability template
+      | From     | Until       | Days     | TimeFrom | TimeUntil | SlotLength | Capacity | Services |
+      | Tomorrow | Tomorrow_+2 | Relative | 09:00    | 10:00     | 5          | 1        | COVID    |
+    Then the request is successful and the following daily availability is created
+      | Date        | From  | Until | Services | Slot Length | Capacity |
+      | Tomorrow    | 09:00 | 10:00 | COVID    | 5           | 1        |
+      | Tomorrow_+1 | 09:00 | 10:00 | COVID    | 5           | 1        |
```
vicr1, line 11: How much work would it be to have more natural language here such as `2 days from now` instead of Tomorrow_+1

vicr1 (dismissed):

tests/Nhs.Appointments.Api.Integration/BaseFeatureStepsTests.cs
```diff
@@ -0,0 +1,30 @@
...
+    [InlineData("Today", 0)]
+    [InlineData("Tomorrow", 1)]
+    [InlineData("Yesterday", -1)]
+    [InlineData("3 days from now", 3)]
+    [InlineData("3 days from today", 3)]
+    [InlineData("3 days before now", -3)]
+    [InlineData("1 day before now", -1)]
+    [InlineData("1 week from now", 7)]
+    [InlineData("2 weeks from today", 14)]
+    [InlineData("1 month before now", -28)]
```
vicr1, line 20: Is this correct? I expect one month before now to be the 20th October which is 31 days

vicr1 (approved):

PR #288, opened 2024-11-20
Appt-62 - Add permission check on Create availability card

ste-banks:


src/new-client/src/app/site/[site]/site-page.tsx
```diff
@@ -22,14 +18,16 @@ export const SitePage = ({ site, permissions }: SitePageProps) => {
         </div>
       </div>
       <ul className="nhsuk-grid-row nhsuk-card-group">
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/site-page.tsx:20]: There used to be a check here to avoid rendering the `<ul>` element if there aren't any `<li>`s to show within it. This would occur if the user had none of the "permissionsRelevantToCards" for desperate want of a better term :D 
Could we add this back in? Otherwise we'll probably get browsers errors about an empty `<ul>`

src/new-client/src/app/site/[site]/site-page.test.tsx
```diff
@@ -75,9 +75,7 @@ describe('Site Page', () => {
     ).not.toBeInTheDocument();
   });
 
-  // TODO: Maybe parameterise these tests over permission/card pairs
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/site-page.test.tsx:78]: How come we decided not to parameterise these?
I don't feel strongly, just curious

src/new-client/src/app/site/[site]/site-page.tsx
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/site-page.tsx:20]: Yeah I can add that in, it wasn't there when I made these changes mind you. They were only on the `<li>` elements so the `<ul>` would have been rendered regardless.

Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/site-page.tsx:20]: Yeah, I removed it when this card was added without permissions because at that time we could guarantee there'd always be at least 1 element in the list. I probably should've replaced it with a TODO tbf rather than just removing it temporarily

sambiramairelogic (dismissed):

sambiramairelogic (approved):

PR #294, opened 2024-11-22
APPT-XXXX: Seed cosmos db before each Playwright test run

sambiramairelogic:
We talked ages ago about a setup/teardown sequence for the playwright tests. 

As we start to write more and more complex tests (such as ones created/editing availability) I think we're going to need patterns for this. 

This is in no means the final solution, but a step towards. This creates a global setup script which re-creates the effect of the .NET cosmos db seeder, running the json files in before a run. 

It does this ONCE before the WHOLE run, rather than before each test file or even before each individual test. We may want to run it between test suites later on.


src/new-client/testing/global-setup.ts
```diff
@@ -0,0 +1,55 @@
...
+import { CosmosClient } from '@azure/cosmos';
+import { type FullConfig } from '@playwright/test';
+
+const { COSMOS_ENDPOINT, COSMOS_TOKEN, SEED_COSMOS_BEFORE_RUN } = env;
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+async function globalSetup(config: FullConfig) {
+  if (SEED_COSMOS_BEFORE_RUN) {
+    await seedCosmos();
+  }
```
sambiramairelogic, line 15: @kicr1 I put in this config value because I know the pipeline runs the .NET version so we may not want to run it again here

kicr1 (approved):

PR #295, opened 2024-11-22
Appt 448 appointment status field

vicr1:
Remove the provisional field from booking, instead use the status field (renamed from outcome) - this can now store the status of an appointment as either Provisional, Booked, or Cancelled. This data is stored in the booking document and booking_index document. Changes also made to the Make Booking endpoint - now accepts "kind" instead of "provisional" - kind can be "booked" or "provisional"

### DIFF
--- a/src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
+++ b/src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
@@ -22,9 +22,11 @@ public class AvailabilityCalculator(IAvailabilityStore availabilityStore, IBooki
 
             var bookings = await bookingDocumentStore.GetInDateRangeAsync(from.ToDateTime(new TimeOnly(0, 0)), until.ToDateTime(new TimeOnly(23, 59)), site);
 
-            var isNotCancelled = (Booking b) => b.Outcome?.ToLower() != "cancelled";
-            var isNotExpiredProvisional = (Booking b) => b.Provisional == false || b.Created.AddMinutes(5) > time.GetUtcNow();
-            var liveBookings = bookings.Where(isNotCancelled).Where(isNotExpiredProvisional);
+            var liveStatuses = new[] { AppointmentStatus.Booked, AppointmentStatus.Provisional };
+            var isExpiredProvisional = (Booking b) => b.Status == AppointmentStatus.Provisional && b.Created < time.GetUtcNow().AddMinutes(-5);
+            var liveBookings = bookings
+                .Where(b => liveStatuses.Contains(b.Status))
+                .Where(b => isExpiredProvisional(b) == false);
 
             foreach (var booking in liveBookings)
             {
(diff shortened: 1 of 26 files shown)


src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
```diff
@@ -22,9 +22,11 @@ public async Task<IEnumerable<SessionInstance>> CalculateAvailability(string sit
 
             var bookings = await bookingDocumentStore.GetInDateRangeAsync(from.ToDateTime(new TimeOnly(0, 0)), until.ToDateTime(new TimeOnly(23, 59)), site);
 
-            var isNotCancelled = (Booking b) => b.Outcome?.ToLower() != "cancelled";
-            var isNotExpiredProvisional = (Booking b) => b.Provisional == false || b.Created.AddMinutes(5) > time.GetUtcNow();
-            var liveBookings = bookings.Where(isNotCancelled).Where(isNotExpiredProvisional);
+            var liveStatuses = new[] { AppointmentStatus.Booked, AppointmentStatus.Provisional };
```
sambiramairelogic, line 25: This is probably just because I've not had my head in the bookings code as much, so feel free to ignore me if all this makes sense, but I just wanted to query whether "live" was the right term here. It feels odd to me that a provisional booking would be considered live

vicr1: I think a provisional booking is potentially live as it is affecting availability. Happy to adopt a different name, but nothing better springs to mind

src/api/Nhs.Appointments.Core/BookingsService.cs
```diff
@@ -59,16 +59,16 @@ public Task<IEnumerable<Booking>> GetBookingByNhsNumber(string nhsNumber)
                 booking.ReminderSent = false;
                 await bookingDocumentStore.InsertAsync(booking);
 
-                if (!booking.Provisional)
+                if (booking.Status == AppointmentStatus.Booked && booking.ContactDetails?.Length > 0)
```
sambiramairelogic, line 62: This confused me at first because booking.Status comes from outside the method and is determined before `canBook` is calculated - I thought "surely the booking can't be booked before we've asked if it can be booked". 

I realise now that this is all coming down from the fact that the consumer themselves has to specify status is Booked when they invoke the endpoint - it feels a bit redundant that they'd have to specify this when trying to make a booking. Feels a bit like `ChangeName(newName: "Fred", nameStatus: Changed)`

I guess the alternative is to create two separate endpoints for `MakeBooking` and `MakeProvisionalBooking`, or find some way of them specifying a `Provisional` flag without having to provide the `Booked` flag in its stead. Maybe the term "book" is just a little overloaded here, and we could separate it out into the concept of "Confirming a booking" vs "Booking a booked booking"

PR #296, opened 2024-11-25
Appt 133 - Add essential docs to cosmos as part of deployment

kicr1:
- Rewrite cosmos scripts to support other cosmos instances and secret vars
- Rename mock-data folder to data/ and add env support for different docs
- Update seeder to support environments and behaviors based on environment that is being seeded
- Add tasks to deployment template for seeding data to environment being deployed


src/new-client/playwright.config.ts
```diff
@@ -21,7 +21,7 @@ export default defineConfig({
   // Let's try this for a bit with the goal of eliminating flakiness, but add in retries if we have to.
   retries: process.env.CI ? 2 : 0,
   // TODO: Playwright init defaults to opt out of parallel tests on CI. We should confirm if we agree with that
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 4 : undefined,
```
sambiramairelogic, line 24: Ooh good shout addressing this - hopefully might speed things up a bit! Could we remove the comment above it too if we think it's now resolved? 

Also, is there any particular reason for landing on 4? According to [the docs](https://playwright.dev/docs/api/class-testconfig#test-config-workers), if we leave this undefined it'll default to 50% of available CPU cores. We could even whack it right up to `'100%'` or 80% or something and see how it copes?

src/new-client/testing/log-in.spec.ts
sambiramairelogic, line 1: Thanks for fixing my silly warnings, just noticed I'd let them slip into main >.< mia culpa for sure

sambiramairelogic (approved):

PR #300, opened 2024-11-26
Appt 337/view weekly availability - PART 1

ste-banks:
This is part 1 of 2 pull requests for the view weekly availability view. This part includes:

- Month view with a card for each week
- Booked appointment counts per service in each week (0 if no booked appointments for that week)
- Logic to work all of the above out (including weeks starting in the month before and ending in the next month)
- Total appointments, total booked count, total un-booked count
- View week link (doesn't currently go anywhere)


src/new-client/src/app/lib/services/viewAvailabilityService.ts
```diff
@@ -0,0 +1,181 @@
...
+    end = start === 1 && end === 8 ? 1 : end;
+    if (end > numDays) {
+      end = numDays;
+    }
+  }
+
+  return weeks;
+};
+
+export const getWeeksInMonth = (year: number, month: number): Week[] => {
```
sambiramairelogic, line 51: We're fairly committed to using DayJS as our datetime handler elsewhere in the code, so it might be nice to be consistent here rather than using JS' own date object. 

Please ignore this if not helpful, but you might actually be able to adapt some similar methods that were also written for a [previous prototype](src/new-client/src/app/lib/services/timeService.ts), e.g. 
image

src/new-client/src/app/lib/services/viewAvailabilityService.ts
```diff
@@ -0,0 +1,181 @@
...
+): Promise<Week[]> => {
+  const fromDate = new Date(
+    weeks[0].startYear,
+    weeks[0].startMonth,
+    weeks[0].start,
+    0,
+    0,
+    0,
+  );
+  const toDate = new Date(
```
sambiramairelogic, line 141: Same as above comment, I'd rather switch over to using DayJS for the sake of consistency

src/new-client/src/app/lib/services/appointmentsService.ts
```diff
@@ -298,3 +302,21 @@ export const setSiteInformationForCitizen = async (
   handleEmptyResponse(response);
   revalidatePath(`/site/${site}/details`);
 };
+
+export const fetchAvailability = async (payload: FetchAvailabilityRequest) => {
```
sambiramairelogic, line 306: Might be okay for now, but it's worth noting that NextJS' only automatically memoizes GET requests because usually a POST would cause changes. If we ever find ourselves calling this more than once in a component tree we should consider adding in a `cache()` call

https://nextjs.org/docs/app/building-your-application/caching#request-memoization

https://nextjs.org/docs/app/building-your-application/caching#react-cache-function

src/new-client/src/app/lib/services/viewAvailabilityService.ts
```diff
@@ -0,0 +1,181 @@
...
+    weeks[weeks.length - 1].endYear,
+    weeks[weeks.length - 1].endMonth,
+    weeks[weeks.length - 1].end,
+    23,
+    59,
+    59,
+  );
+
+  console.log(fromDate.getFullYear(), toDate.getFullYear());
```
sambiramairelogic, line 150: Console log left in :D

src/new-client/src/app/lib/services/viewAvailabilityService.ts
```diff
@@ -0,0 +1,181 @@
...
+    const countArray = blocks.map(b => b.count);
+    unbookedCount =
+      countArray.length === 0 ? 0 : countArray.reduce((sum, num) => sum + num);
+  }
+
+  return unbookedCount;
+};
+
+const weekStart = (week: Week): Date => {
+  return new Date(week.startYear, week.startMonth, week.start, 0, 0, 0);
```
sambiramairelogic, line 108: We might be able to save ourselves a lot of manual maths if we switch to DayJs, e.g. the [isoWeek plugin](https://day.js.org/docs/en/plugin/iso-week) would let us just go `date.startOf('isoWeek')`. Same goes for start of month/year etc. Might we be able to simplify the new `Week` type a little too, as we might not need the `monthStart`/`yearStart` properties for example?

src/new-client/src/app/lib/services/appointmentsService.ts
sambiramairelogic, line 1: Not sure if this is coming in part 2, but it might be worth resolving these comments now?
image

ste-banks: Hadn't actually noticed those, I can take a look as part of this PR

src/new-client/src/app/lib/types/index.tsx
```diff
@@ -126,8 +126,103 @@ type Session = {
...
+  from: string;
+  duration: number;
+  service: string;
+  site: string;
+  outcome?: string;
+  attendeeDetails: AttendeeDetails;
+  contactDetails?: ContactItem[];
+  reminderSet: boolean;
+  created: string;
+  provisional: boolean;
```
sambiramairelogic, line 186: @vicr1 has just changed this property on the backend to be a new type object describing booked/provisional as a status, so might need updating here

src/new-client/src/app/lib/types/index.tsx
```diff
@@ -126,8 +126,103 @@ type Session = {
...
+  lastName: string;
+  dateOfBirth: Date;
+};
+
+type ContactItem = {
+  type: string;
+  value: string;
+};
+
+type bookedAppointments = {
```
sambiramairelogic, line 202: Sorry silly little nitpick but could this be PascalCase like other type names?

src/new-client/src/app/lib/types/index.tsx
```diff
@@ -126,8 +126,103 @@ type Session = {
   services: string[];
 };
 
+type FetchAvailabilityRequest = {
+  sites: string[];
+  service: string;
+  from: string;
+  until: string;
+  queryType: string;
```
sambiramairelogic, line 134: We could consider locking this down to `"NotSet" | "Days" | "Hours" | "Slots"` for extra type safety and/or readability

src/new-client/src/app/lib/types/index.tsx
ste-banks: Yep, must have been a typo.

src/new-client/src/app/site/[site]/view-availability/view-availability-page.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <Table
+            headers={[
+              `Total appointments: ${week.totalAppointments}`,
+              `Booked: ${week.booked}`,
+              `Unbooked: ${week.unbooked}`,
+            ]}
+            rows={[]}
+          ></Table>
+          <br />
+          <Link href="" className="nhsuk-link">
```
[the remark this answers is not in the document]
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/view-availability/view-availability-page.tsx:33]: I was going to add that in part 2 because the screen it's going to link to doesn't yet exist.

src/new-client/src/app/site/[site]/view-availability/page.tsx
```diff
@@ -0,0 +1,59 @@
...
+  getWeeksInMonth,
+} from '@services/viewAvailabilityService';
+
+type PageProps = {
+  params: {
+    site: string;
+  };
+};
+
+const Page = async ({ params }: PageProps) => {
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/view-availability/page.tsx:17]: Could we add a `await assertPermission(site.id, 'availability:query')` to this so the route is protected with the same permissions that govern the card's visibility?

src/new-client/src/app/site/[site]/view-availability/view-availability-page.tsx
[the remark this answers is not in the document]
Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/view-availability/view-availability-page.tsx:33]: Yeah, I tried the do the same thing when I first joined but Vince made some pretty good arguments for being stricter about it. But I don't wanna speak for him... I'm fine with leaving it in personally as long as it's definitely coming in part 2 :)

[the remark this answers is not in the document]
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/view-availability/view-availability-page.tsx:33]: I'll remove it from this one, easy enough to add back in when there's actually a page for it to go to :)

src/new-client/src/app/lib/types/index.tsx
```diff
@@ -126,8 +127,116 @@ type Session = {
...
+  { label: 'COVID 18-74', value: 'COVID:18_74' },
+  { label: 'COVID 5-11 (10 Min)', value: 'COVID:5_11_10' },
+  { label: 'COVID 75+', value: 'COVID:75' },
+  { label: 'FLU 18-64', value: 'FLU:18_64' },
+  { label: 'FLU 65+', value: 'FLU:65' },
+  { label: 'FLU and COVID 18-64', value: 'COVID_FLU:18_64' },
+  { label: 'FLU and COVID 65-74', value: 'COVID_FLU:65-74' },
+  { label: 'FLU and COVID 75+', value: 'COVID_FLU:75' },
+  { label: 'RSV (Adult)', value: 'RSV:Adult' },
+];
```
sambiramairelogic, line 239: Ooh I really should've spotted this before, apologies. 
This will have a knock-on effect on the Create Availability Wizard, because the list of services you're allowed to select in `select-services-step.tsx` just maps over this array. 
Can we quickly check with the team whether this is fine, or whether we need to filtered it down to just RSV for the beta?

sambiramairelogic (approved):

sambiramairelogic (approved):

PR #301, opened 2024-11-26
APPT-XXX: Resolve .NET warnings in Appointments.Core.UnitTests

sambiramairelogic:
I'd like to turn on TreatWarningsAsErrors (and enforce some auto formatting rules). 
If we did that at the moment we'd also have to make hundreds of changes just to get a passing build, so we need to start resolving some of those warnings beforehand. 
This PR just addresses the warnings in Appointments.Core.UnitTests

### DIFF
--- a/tests/Nhs.Appointments.Core.UnitTests/BookingServiceTests.cs
+++ b/tests/Nhs.Appointments.Core.UnitTests/BookingServiceTests.cs
@@ -230,7 +230,7 @@ namespace Nhs.Appointments.Core.UnitTests
         [Fact]
         public async Task CancelBooking_ReturnsNonSuccess_WhenInvalidReference()
         {
-            _bookingsDocumentStore.Setup(x => x.GetByReferenceOrDefaultAsync(It.IsAny<string>())).Returns(Task.FromResult<Booking?>(null));
+            _bookingsDocumentStore.Setup(x => x.GetByReferenceOrDefaultAsync(It.IsAny<string>())).Returns(Task.FromResult<Booking>(null));
             var result = await _bookingsService.CancelBooking("some-reference");
             Assert.Equal(BookingCancellationResult.NotFound, result);
         }
@@ -241,7 +241,7 @@ namespace Nhs.Appointments.Core.UnitTests
             var site = "some-site";
             var bookingRef = "some-booking";
             
-            _bookingsDocumentStore.Setup(x => x.GetByReferenceOrDefaultAsync(It.IsAny<string>())).Returns(Task.FromResult<Booking?>(new Booking() { Site = site, ContactDetails = [] }));
+            _bookingsDocumentStore.Setup(x => x.GetByReferenceOrDefaultAsync(It.IsAny<string>())).Returns(Task.FromResult<Booking>(new Booking() { Site = site, ContactDetails = [] }));
             _bookingsDocumentStore.Setup(x => x.UpdateStatus(bookingRef, AppointmentStatus.Cancelled)).ReturnsAsync(true).Verifiable();
 
             await _bookingsService.CancelBooking(bookingRef);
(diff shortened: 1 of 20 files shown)


src/api/Nhs.Appointments.Core/AvailabilityService.cs
```diff
@@ -2,7 +2,7 @@
 
 public class AvailabilityService(IAvailabilityStore availabilityStore, IAvailabilityCreatedEventStore availabilityCreatedEventStore) : IAvailabilityService
 {
-    public async Task ApplyAvailabilityTemplateAsync(string site, DateOnly from, DateOnly until, Template template, string user)
+    public async Task ApplyAvailabilityTemplateAsync(string? site, DateOnly from, DateOnly until, Template? template, string user)
```
sambiramairelogic, line 5: These sorts of warnings I'd rather resolve by turning on TreatWarningsAsErrors then leaving the type non-nullable, but at the moment that would cause hundreds of changes in the codebase. 

We're checking for null a few lines down, so to avoid a warning the type currently needs to declare itself as potentially null. I think it all comes down to making the intention clear in the method signature

tests/Nhs.Appointments.Core.UnitTests/BookingServiceTests.cs
```diff
@@ -25,7 +25,7 @@ public BookingServiceTests()
                 TimeProvider.System);
         }
 
-        [Fact (Skip = "flakey test. needs investigation")]
+        [Fact]
```
sambiramairelogic, line 28: Maybe this test still wants to be skipped - I've just addressed the warnings about `task.Wait()` in a fairly naïve way but it's passed 5 times locally and twice on the pipeline, so that might've fixed the flakiness? Maybe? 🤷

ste-banks (dismissed):

src/api/Nhs.Appointments.Core/AvailabilityService.cs
vicr1: I'm not a fan of this I would prefer to turn off nullable reference types

tests/Nhs.Appointments.Core.UnitTests/UserServiceTests.cs
```diff
@@ -89,14 +89,14 @@ public async void IncludesUserIdInEvent()
         }
 
         [Fact]
-        public async void ReturnsFailureWhenRoleNotFound()
+        public async Task ReturnsFailureWhenRoleNotFound()
```
vicr1, line 92: What's the advantage of this?

sambiramairelogic: "Support for 'async void' unit tests is being removed from xUnit. net v3" is the warning this generates - I haven't really looked into exactly why but I was lead to believe that `async Task` is preferable to `async void` anyway.

From a very quick cursory Google it sounds like there are differences in how exceptions are handled in each case, and the test framework might struggle to observe when an `async void` method has completed. I can look into it further if you like

kicr1 (approved):

kicr1:
Is there any reason why we have left nullable enabled on the other test projects within Nhs.Appointments.Api.UnitTests?

sambiramairelogic:
> Is there any reason why we have left nullable enabled on the other test projects within Nhs.Appointments.Api.UnitTests?

I'm not sure I understand - `Nhs.Appointments.Api.UnitTests` is itself a test project. Do you mean the other test project adjacent to it within the /tests directory? I so then because at the advent of this PR I wanted to keep the changes small and reviewable. I was going to do one PR per project until eventually the whole solution is warning free, then ask the team about enabling TreatWarningsAsErrors. 

I'm happy to expand this to other projects if you like?

PR #305, opened 2024-11-27
APPT-29 Fix the open api doc

vicr1:
Add code to rewrite parts of the document that are generated incorrectly


src/api/Nhs.Appointments.Api/Models/TypeFixDocumentFilter.cs
```diff
@@ -0,0 +1,98 @@
...
+                Default = new OpenApiString(time.GetLocalNow().ToString("yyyy-MM-dd"))
+            };
+        }
+        else if(underlyingType == typeof(TimeOnly))
+        {
+            return new OpenApiSchema
+            {
+                Type = "string",
+                Format = "time",
+                Default = new OpenApiString("09:00")
```
sambiramairelogic, line 69: Perhaps this is going too far, but I could we somehow communicate here that this is a 24 hour time? I thought about switching it to 17:00 or something, but then it's not clear if 9:00 is valid as opposed to 09:00. 

Looks like the `OpenApiSchema` type accepts a string `Description` field, perhaps we could add some text in there?

src/api/Nhs.Appointments.Api/Models/TypeFixDocumentFilter.cs
```diff
@@ -0,0 +1,98 @@
...
+    {        
+        var underlyingType = Nullable.GetUnderlyingType(type) ?? type;
+
+        if(underlyingType.IsEnum)
+        {
+            return new OpenApiSchema
+            {
+                Type = "string",
+                Enum = type.GetEnumNames().Select(n => new OpenApiString(n)).Cast<IOpenApiAny>().ToList(),
+                Default = new OpenApiString(type.GetEnumNames().Last())
```
sambiramairelogic, line 51: I'm absolutely fine with it, but out of curiousity why have we chosen the default to be the last here? Just surprised me since when newing up an enum 0 would be the default

vicr1: In some paradigms the default for an enum will (and arguably should) be undefined. This is so that they don't have a meaningful value if they are not set. An example of this is that QueryType the default (0) value is actually NotSet. To prevent these from showing I wanted to use a value other than the default. I chose last to prevent having to test for weird edge cases where there was only one value in the enum

src/api/Nhs.Appointments.Api/Models/TypeFixDocumentFilter.cs
vicr1: Done

PR #308, opened 2024-11-29
Appt 337/view availability pagination - PART 2

ste-banks:
Part 2 includes:
- Pagination
- Logic fixups as the first day of each week was missing from booking count
- Updated tests
- Little bit of code clean up

### DIFF
--- a/src/new-client/src/app/lib/types/index.tsx
+++ b/src/new-client/src/app/lib/types/index.tsx
@@ -180,7 +180,7 @@ type Booking = {
   duration: number;
   service: string;
   site: string;
-  status: AppointmentStatus;
+  status: 'Unknown' | 'Provisional' | 'Booked' | 'Cancelled';
   attendeeDetails: AttendeeDetails;
   contactDetails?: ContactItem[];
   reminderSet: boolean;
@@ -196,7 +196,7 @@ type AttendeeDetails = {
 };
 
 type ContactItem = {
-  type: ContactItemType;
+  type: 'Phone' | 'Email' | 'Landline';
   value: string;
 };
 
@@ -209,19 +209,6 @@ type ClinicalService = {
   label: string;
   value: string;
 };
-
-enum AppointmentStatus {
-  Unknown = 'Unknown',
-  Provisional = 'Provisional',
-  Booked = 'Booked',
-  Cancelled = 'Cancelled',
-}
-
-enum ContactItemType {
-  Phone,
-  Email,
-  Landline,
-}
 
 // TODO: Decide where this info should live and move it there
 const clinicalServices: ClinicalService[] = [
@@ -259,11 +246,4 @@ export type {
   Week,
 };
 
-export {
-  AppointmentStatus,
-  ContactItemType,
-  MyaError,
-  UnauthorizedError,
-  daysOfTheWeek,
-  clinicalServices,
-};
+export { MyaError, UnauthorizedError, daysOfTheWeek, clinicalServices };
--- a/src/new-client/src/app/lib/types/index.tsx
+++ b/src/new-client/src/app/lib/types/index.tsx
@@ -211,10 +211,10 @@ type ClinicalService = {
 };
 
 enum AppointmentStatus {
-  Unknown,
-  Provisional,
-  Booked,
-  Cancelled,
+  Unknown = 'Unknown',
+  Provisional = 'Provisional',
+  Booked = 'Booked',
+  Cancelled = 'Cancelled',
 }
 
 enum ContactItemType {
--- a/src/new-client/src/app/lib/types/index.tsx
+++ b/src/new-client/src/app/lib/types/index.tsx
@@ -164,8 +164,8 @@ type Week = {
   totalAppointments?: number;
   booked?: number;
   bookedAppointments: BookedAppointments[];
-  startDate?: dayjs.Dayjs;
-  endDate?: dayjs.Dayjs;
+  startDate: dayjs.Dayjs;
+  endDate: dayjs.Dayjs;
 };
 
 type FetchBookingsRequest = {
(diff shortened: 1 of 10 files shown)


src/new-client/src/app/site/[site]/view-availability/page.tsx
```diff
@@ -10,25 +10,33 @@ import { FetchAvailabilityRequest } from '@types';
 import {
   getDetailedMonthView,
   getWeeksInMonth,
-  monthEnd,
-  monthStart,
 } from '@services/viewAvailabilityService';
+import customParseFormat from 'dayjs/plugin/customParseFormat';
+
+dayjs.extend(customParseFormat);
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/view-availability/page.tsx:16]: There's already a method for this in the `timeService` file, so you don't need to extend with the plugin in multiple files: 

`export const parseDateString = (dateString: string, format = 'YYYY-MM-DD') => {
  return dayjs.utc(dateString, format, true);
};`

Ideally, I'd have liked all our usages of a third party library to be behind an abstraction so that if we change to a different library at some point it's only one file. Realistically though, that'd just mean creating hundreds of wrapper methods so we have ended up importing dayJs all over the place. 

Because we need the plugin extension for this one I thought it was a good candidate for being given an abstracted wrapper and avoid importing this plugin in multiple files.

src/new-client/src/app/site/[site]/view-availability/week/page.tsx
```diff
@@ -0,0 +1,22 @@
+import NhsPage from '@components/nhs-page';
+import dayjs from 'dayjs';
+
+type PageProps = {
+  searchParams: {
+    from: string;
+    until: string;
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/view-availability/week/page.tsx:7]: What's the need for the `until` parameter here? I'm just thinking that any searchParams are open to the users fudging them around. They could enter an `until` date that's a year away or in the past so we'll have to do some validation if we go down this route, at which point wouldn't it be simpler just to calculate the week from the week start date? 

Or go even further, and accept any single date and calculate the week even if it isn't the start? So if the user enters today's date that'll see 25th -> 1st? Should be doable with `date.startOf('isoWeek') and `date.endOf(isoWeek)`.

src/new-client/src/app/site/[site]/view-availability/week/page.tsx
```diff
@@ -0,0 +1,22 @@
...
+  };
+};
+
+const Page = async ({ searchParams }: PageProps) => {
+  return (
+    <NhsPage
+      title={`${dayjs(searchParams.from).format('D MMMM')} to ${dayjs(searchParams.until).format('D MMMM')}`}
+      breadcrumbs={[{ name: 'Home', href: '/' }]}
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/view-availability/week/page.tsx:15]: I think we might be missing a breadcrumb here - shouldn't `/site/${params.site}/view-availability` be in here?

src/new-client/src/app/site/[site]/view-availability/week/page.tsx
```diff
@@ -0,0 +1,22 @@
...
+import dayjs from 'dayjs';
+
+type PageProps = {
+  searchParams: {
+    from: string;
+    until: string;
+  };
+};
+
+const Page = async ({ searchParams }: PageProps) => {
```
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/view-availability/week/page.tsx:11]: I think we might need to add `await assertPermission(site.id, 'availability:query');` here, same as its parent page?

src/new-client/src/app/site/[site]/view-availability/week/page.tsx
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/view-availability/week/page.tsx:15]: I left this for now because it would be covered by APPt-338 as it was next on my list. Though admittedly I forgot to add a TODO to revisit it in that ticket. I'll add it in

src/new-client/src/app/site/[site]/view-availability/week/page.tsx
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/view-availability/week/page.tsx:11]: One day I will remember to add that in to a new page :D

src/new-client/src/app/site/[site]/view-availability/week/page.tsx
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/view-availability/week/page.tsx:7]: It was just for the title of the new screen. Happy to update it but it's probably worth addressing in APPT-338 which covers this new page. I only added it in so the link went somewhere. I think I'll add TODOs in this file for now to be covered when either of us pick up APPT-338 :)

src/new-client/src/app/site/[site]/view-availability/week/page.tsx
Comment (sambiramairelogic) [reply src/new-client/src/app/site/[site]/view-availability/week/page.tsx:11]: I did consider making it a required parameter of `<NhsPage/>` and `<NhsTransactionalPage/>` for just that reason :D Feel free to do so, I only held off cos I wasn't sure if there'd be pages with an AND/OR condition. Pretty sure atm though it's just one role or none so would be trivial

src/new-client/src/app/site/[site]/view-availability/page.tsx
Comment (ste-banks) [reply src/new-client/src/app/site/[site]/view-availability/page.tsx:16]: Turns out I had used this previously then removed it but didn't remove the import.

src/new-client/src/app/lib/types/index.tsx
```diff
@@ -211,10 +211,10 @@ type ClinicalService = {
...
 
 enum AppointmentStatus {
-  Unknown,
-  Provisional,
-  Booked,
-  Cancelled,
+  Unknown = 'Unknown',
+  Provisional = 'Provisional',
+  Booked = 'Booked',
+  Cancelled = 'Cancelled',
```
sambiramairelogic, line 217: Are these strings needed because otherwise `AppointmentStatus.Provisional` is coming out as `1`? I'd argue at this point it'd be safer to just convert it to a union type, then there's chance for confusion or differences in compiler/runtime behaviour. 

If it stays as an enum, wouldn't the same strings be required for `ContactItemType`?

sambiramairelogic (approved):

PR #314, opened 2024-11-29
Remove left over file from old client removal

kicr1:


src/new-client/src/app/site/[site]/view-availability/week/page.tsx
Comment (sambiramairelogic) [inline src/new-client/src/app/site/[site]/view-availability/week/page.tsx:1]: Oh damn, this must've been a rebase mistake, my bad. 

I think actually this wants moving rather than deleting, it's part of Ste's in-flight work. It should be in: `...src\app\site\[site]\view-availability\week\page.tsx`

PR #316, opened 2024-11-29
APPT-258 added tests for E2E single/weekly create availability wizard

sarithalakkyreddy:


src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
...
+    await this.page.getByLabel('Session start time - hour').fill(hour);
+    await this.page.getByLabel('Session start time - minute').fill(minute);
+  }
+
+  async enterEndtTime(hour: string, minute: string) {
+    await this.page.getByLabel('Session end time - hour').fill(hour);
+    await this.page.getByLabel('Session end time - minute').fill(minute);
+  }
+
+  async noOfVaccinators(numberof: string) {
```
sambiramairelogic, line 88: It's convention in JavaScript/Typescript to write method names and parameters in camelCase, so this should be `numberOf` rather than `numberof`

src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
...
+    await endDateFormGroup.getByLabel('Month').fill(month);
+    await endDateFormGroup.getByLabel('Year').fill(year);
+  }
+
+  async enterStartTime(hour: string, minute: string) {
+    await this.page.getByLabel('Session start time - hour').fill(hour);
+    await this.page.getByLabel('Session start time - minute').fill(minute);
+  }
+
+  async enterEndtTime(hour: string, minute: string) {
```
sambiramairelogic, line 83: There looks to be a typo here, this should be `enterEndTime`

src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
...
+
+  async appointmentLength(length: string) {
+    await this.page.getByLabel('How long are your appointments?').fill(length);
+  }
+
+  async addServices(serviceName: string) {
+    await this.page.getByRole('checkbox', { name: serviceName }).click();
+  }
+
+  async selectDay(Day: string) {
```
sambiramairelogic, line 102: As mentioned above, this should be `day` rather than `Day`

src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
...
+    await this.page
+      .getByLabel('How many vaccinators or vaccination spaces do you have?')
+      .fill(numberof);
+  }
+
+  async appointmentLength(length: string) {
+    await this.page.getByLabel('How long are your appointments?').fill(length);
+  }
+
+  async addServices(serviceName: string) {
```
sambiramairelogic, line 98: This method name is plural, but only adds a singular service. It might be clearer if it was called just `addService`

src/client/testing/page-objects/create-availability.ts
sambiramairelogic: Also all the other method names are verbs (i.e. `enterSomething` / `selectSomething`) but this one isn't, perhaps it should be `enterNoOfVaccinators` ?

src/client/src/app/site/[site]/create-availability/wizard/days-of-week-step.tsx
```diff
@@ -99,6 +99,7 @@ const DaysOfWeekStep = ({
           <CheckBox
             label={'Select all days'}
             value={daysOfTheWeek}
+            id={'alldays'}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/create-availability/wizard/days-of-week-step.tsx:102]: Could we make this camelCase too to be consistent? i.e. `allDays`

src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
...
+
+    this.btnCreateAvailability = page.getByRole('button', {
+      name: 'Create availability',
+    });
+
+    this.sessionTittle = page.getByRole('heading', {
+      name: 'What type of session do you want to create?',
+    });
+
+    this.btnContinue = page.getByRole('button', {
```
sambiramairelogic, line 27: This is a very minor thing and just my opinion, so feel free to ignore me if you want, but in the other page object models we're trying to describe components the way you would in a real life sentence, for example `logOutButton`. I'd personally rather stick to this pattern rather than start adding `btn` to the start. 

Personally I find `continueButton.click()` easier to read than `btnContinue.click()`, but like I say that's just my opinion! 

It'd also be nice to keep things consistent, so if we do decide to go down this route then we should rename the other models in kind (i.e. change `logOutButton` to `btnLogOut` )

src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
 
 export default class CreateAvailabilityPage extends RootPage {
   readonly title: Locator;
+  readonly btnCreateAvailability: Locator;
+  readonly sessionTittle: Locator;
```
sambiramairelogic, line 7: Type here - tittle vs title

src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
...
+  readonly btnContinue: Locator;
+  readonly setAndCapacityTittle: Locator;
```
sambiramairelogic, line 9: and here

src/client/testing/create-availability.spec.ts
```diff
@@ -31,3 +31,53 @@ test('A user can navigate to the Create Availability flow from the site page', a
   await page.waitForURL('**/site/ABC02/create-availability');
   await expect(createAvailabilityPage.title).toBeVisible();
 });
+
+test('E2E Create single session of RSV availability', async ({ page }) => {
```
sambiramairelogic, line 35: Again it's a really minor thing and just my opinion, but I don't think you need to write `E2E` at the start of your test names. The other tests don't so it's a bit inconsistent, and since these are all end-to-end tests anyway it doesn't tell us anything we don't already know. 

If it's a pattern you want to introduce however, we should also add `E2E` to the start of all the other test names for consistency :)

sarithalakkyreddy:
@sambiramairelogic Review comments update please take a look.

src/client/testing/page-objects/create-availability.ts
```diff
@@ -3,11 +3,103 @@ import RootPage from './root';
 
 export default class CreateAvailabilityPage extends RootPage {
   readonly title: Locator;
+  readonly CreateAvailabilityButton: Locator;
```
sambiramairelogic, line 6: This should be camelCase - i.e. `createAvailabilityButton`

src/client/testing/create-availability.spec.ts
```diff
@@ -31,3 +31,53 @@ test('A user can navigate to the Create Availability flow from the site page', a
...
+  await createAvailabilityPage.enterSingleDateSessionDate('27', '10', '2025');
+  await createAvailabilityPage.continueButton.click();
+  await createAvailabilityPage.enterStartTime('09', '00');
+  await createAvailabilityPage.enterEndTime('10', '00');
+  await createAvailabilityPage.enterNoOfVaccinators('1');
+  await createAvailabilityPage.appointmentLength('5');
+  await createAvailabilityPage.continueButton.click();
+  await createAvailabilityPage.addService('RSV (Adult)');
+  await createAvailabilityPage.continueButton.click();
+  await createAvailabilityPage.btnSaveSession.click();
```
sambiramairelogic, line 55: Could this be `saveSessionButton` also to match the others?

sarithalakkyreddy:
Its updated now take a look

sambiramairelogic (dismissed):

sambiramairelogic (approved):

PR #320, opened 2024-12-02
first pass at MYA site csv to json tool

jrowett:


data/CsvDataTool/SiteCsvReader.cs
```diff
@@ -0,0 +1,89 @@
...
+                sites.Add(site);
+                report.Add(new SiteRowReportItem(index, site.Name, true, ""));
+                index++;
+            }       
+        }
+
+        return (sites.ToArray(), report.ToArray());
+    }
+
+    private TextReader GetCsvStreamReader()
```
vicr1, line 75: This could be replaced by using the Lazy pattern

data/CsvDataTool/SiteJsonWriter.cs
vicr1, line 1: Seems like a potential gotcha having this duplicated

data/CsvDataTool/SiteMap.cs
```diff
@@ -0,0 +1,52 @@
...
+            new Location(
+                "Point",
+                [x.Row.GetField<double>("Longitude"), x.Row.GetField<double>("Latitude")]
+                ));
+        Map(m => m.IntegratedCareBoard).Name("ICB");
+        Map(m => m.Region).Name("Region");
+    }
+}
+
+public record struct Site(
```
vicr1, line 24: Can we reference the existing types

vicr1 (changes requested):
Should be marked as draft as this is not a complete piece of work

data/CsvDataTool/SiteCsvReader.cs
```diff
@@ -0,0 +1,89 @@
...
+            Quote = '"'
+        };
+
+        using (var reader = GetCsvStreamReader())
+        using (var csv = new CsvReader(reader, config))
+        {
+            csv.Context.RegisterClassMap<SiteMap>();
+
+            var imported = csv.GetRecords<Site>();
+            foreach(var site in imported)
```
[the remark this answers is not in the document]
jrowett: i don't see how this increments the value of index

data/CsvDataTool/SiteMap.cs
jrowett: the CSV library has slightly different requirements for the DTOs, they need default constructors. this was my original approach until i encountered this gotcha in the library

data/CsvDataTool/SiteJsonWriter.cs
jrowett: puts s dependency from this tool onto the API function app - we could cleave off json tasks to their own library i guess

vicr1 (approved):

PR #344, opened 2024-12-06
Fix for sonar issue redirect on client input

vicr1:


src/api/Nhs.Appointments.Api/Auth/ServiceRegistration.cs
vicr1, line 1: Change to this file is swap to top level namespace declaration has caused git to get confused with the diff. Main highlight is the addition of a new property when configuring AuthOptions

kicr1 (approved):

PR #347, opened 2024-12-09
Feature/appt 384 site details on homepage

Dave-O-NHS:


src/client/src/app/lib/components/nhsuk-frontend/summary-list.tsx
```diff
@@ -14,7 +14,7 @@ type OnClickActionProps = {
 
 export type SummaryListItem = {
   title: string;
-  value: string;
+  value: string | string[];
```
ste-banks, line 17: Instead of making this a `string[]` would it be better if we made it `string | ReactNode` similar to how it is done in the `<Table .. />` component?

vicr1: I am not sure I agree. I think in this case doing the simplest thing that works is fine.

src/client/src/app/site/[site]/site-page.tsx
```diff
@@ -61,3 +58,23 @@ export const SitePage = ({ site, permissions }: SitePageProps) => {
...
+const mapSummaryData = (site: Site) => {
+  if (!site) {
+    return undefined;
+  }
+
+  const items: SummaryListItem[] = [];
+
+  items.push({
+    title: 'Address',
+    value: site.address.match(/[^,]+,|[^,]+$/g) || [], // Match each word followed by a comma, or the last word without a comma
```
Comment (vicr1) [inline src/client/src/app/site/[site]/site-page.tsx:71]: Does split not work here? Seems quite complex implementation

vicr1 (approved):

src/client/src/app/site/[site]/site-page.tsx
Comment (Dave-O-NHS) [reply src/client/src/app/site/[site]/site-page.tsx:71]: split works but it removed comma at the end of each item which goes agains UI design. Got some other ideas how to split by comma by keep it in items?

src/client/src/app/lib/components/nhsuk-frontend/summary-list.tsx
ste-banks: Yeah fair enough, happy to go with that.

PR #348, opened 2024-12-09
Appt 338/daily availability front end - PART 2

ste-banks:
Part 2 of 2 for viewing daily availability. This part includes:

- Logic to populate the daily views using the new API call from part 1 and the existing bookings query endpoint
- Pagination to go to the previous and next weeks
- Further unit tests for the front end changes & new service methods
- TODO comments for features / logic that is related but out of scope for APPT-338


src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx
```diff
@@ -1,35 +1,99 @@
...
+  const nextWeek = weekStart.add(1, 'week');
+  const previousWeek = weekStart.add(-1, 'week');
+
+  let previousWeekDisplay = '';
+  let nextWeekDisplay = '';
+
+  if (previousWeek.month() < weekStart.month()) {
+    previousWeekDisplay = `${previousWeek.format('D MMMM')}-${weekEnd.add(-1, 'week').format('D MMMM YYYY')}`;
+  } else {
+    previousWeekDisplay = `${weekStart.add(-1, 'week').format('D')}-${weekEnd.add(-1, 'week').format('D MMMM YYYY')}`;
```
Comment (vicr1) [inline src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx:25]: Why do we not use previousWeek in place of weekStart.add(-1 'week')? This is also confusing so maybe refactor to a method so that we can document what this is doing in the method name

Comment (ste-banks) [reply src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx:25]: Yeah I should have used previous week for both, I'll clean this up

Dave-O-NHS (approved):
LGTM

PR #370, opened 2024-12-19
Appt 239 display session capacity calculation for a single weekly session

sarithalakkyreddy:
test resulsts


sambiramairelogic (changes requested):
There are several linting errors on this branch that need to be addressed. Before committing code, we should be running the `npm run lint` command from the `~/src/client` folder. Running that on this branch currently yields the following warnings 
image

src/client/testing/utils/date-utility.ts
sambiramairelogic, line 1: It's great that we've abstracted this out, but I think we could improve on a few things here. 

Firstly, in modern Javascript we don't need to bother creating a class to hold these methods. Each file is a module so we can just export methods or variables directly. 

Secondly, we don't need to convert dates to ISO format then perform string manipulations in order to format them - most date libraries already have methods to do this. Using Javascript's own native dates we could do `new Date().toLocaleDateString("en-GB")` which would format the date in the format "01/01/2000". Javascript's native date library is pretty poor though, so elsewhere in this project we're using DayJS. 
Using DayJS, you can provide the exact pattern you want the date to be formatted to, so `dayjs().format('DD/MM/YYYY')` would do the same. 

Thirdly, we should strongly type the return object from this method. At the moment it's returning an array of strings. This isn't obvious to other coders that we actually only expect this array to be of length 3, or that it includes string representations of date components. We could create an interface to describe this better, and make future use clearer. 

Doing those 3 things, we might end up with something a bit more like this: 
image

Because the return type is now an interface, we can access its properties by name rather than the array index so usage is clearer too: 
image

sambiramairelogic: Docs for Day JS -> https://day.js.org/docs/en/display/format

src/client/testing/create-availability.spec.ts
```diff
@@ -3,71 +3,81 @@ import RootPage from './page-objects/root';
...
+let summarypage: SummaryPage;
 
 test.beforeEach(async ({ page }) => {
   rootPage = new RootPage(page);
   oAuthPage = new OAuthLoginPage(page);
   siteSelectionPage = new SiteSelectionPage(page);
   sitePage = new SitePage(page);
   createAvailabilityPage = new CreateAvailabilityPage(page);
-});
+  dateUtils = new DateUtils();
```
sambiramairelogic, line 24: If we make this a module as suggested in my other comments, we won't need to new it up or declare it here

src/client/testing/page-objects/create-availability/summary-page.ts
```diff
@@ -0,0 +1,21 @@
...
+    this.summaryPageTitle = page.getByText('Check single date session');
+
+    this.saveSessionButton = page.getByText('Save session');
+  }
+
+  async changeFunctionalityLink(val: string) {
+    const row = this.page.locator('.nhsuk-summary-list__row', { hasText: val });
+    const changeLink = row.locator('.nhsuk-summary-list__actions a');
+    await changeLink.click();
+  }
```
sambiramairelogic, line 20: We should avoid using classname selectors as this isn't how screen readers interact with the DOM. If we can't find an element using roles / labels / visible text, then there's a good chance screen readers can't either. 

The summary list uses `<dl>` elements which, [according to the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#technical_summary), can be given the role `list`. I think therefore we should make the following changes to `summary-list.tsx`: 
image

and then rewrite this method to be something along the lines of:

image
(that may not work exactly, some tweaking may be required)

sambiramairelogic:
Closing as this work has now been reproduced in https://github.com/NHSDigital/nbs-appointments-management-service/pull/422

PR #382, opened 2024-12-27
Avoid unnecessary server trips when using tabs and pagination

sambiramairelogic:
- Resolves console errors that `tabs.test.tsx` was causing
- Moves the logic for handling tab selection into the Tabs component so the consumer doesn't have to implement it themselves
- Refactors the tabs/pagination on these pages to avoid unnecessary server round-trips

### DIFF
--- a/src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx
+++ b/src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx
@@ -1,5 +1,5 @@
 'use client';
-import { Table, ButtonGroup } from '@nhsuk-frontend-components';
+import { Table, Pagination } from '@nhsuk-frontend-components';
 import { AttendeeDetails, ContactItem, Booking } from '@types';
 import { formatDateTimeToTime, dateToString } from '@services/timeService';
 import { ReactNode } from 'react';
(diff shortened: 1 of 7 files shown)


src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx
```diff
@@ -112,12 +100,35 @@ export const DailyAppointmentsPage = ({
...
   };
+
   const appointmentsTableData = mapTableData();
 
   return (
     <>
       {appointmentsTableData && <Table {...appointmentsTableData}></Table>}
-      <Pagination previous={buildPreviousPage()} next={buildNextPage()} />
+
+      {/* TODO: Implement the pagination component and in the first screenshot of https://nhsd-jira.digital.nhs.uk/browse/APPT-367 and replace these links
```
Comment (vicr1) [inline src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx:110]: Not sure this should be a to do item. The initial version of this page will use the simplified paging components. A future version of this page might look different but that should be as per user feedback from the beta rather than assuming how it should be done

src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
Comment (vicr1) [inline src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:1]: How does the new implementation handle page numbers when the user switches tabs? If I am on page 4 of the active bookings and I switch to cancelled does it reset to page 1?

vicr1 (approved):

src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx
Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx:110]: Good point, when I did this I thought the more complex pagination was definitely queued up for a later ticket. I'll extend the existing pagination component to accept onClick links as well as href ones

src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:1]: Aha, very good spot. I'd missed that. Currently it'll only update the tab parameter and stay on the current page number - will make it reset to 1

PR #391, opened 2025-01-02
Appt 340 additional session availability to a single day

Dave-O-NHS:


src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx
```diff
@@ -76,15 +92,21 @@ const AvailabilityTemplateWizard = ({ site }: Props) => {
...
+          {dateIsNotProvided() && (
+            <WizardStep>
+              {stepProps => <StartAndEndDateStep {...stepProps} />}
+            </WizardStep>
+          )}
+          {dateIsNotProvided() && (
+            <WizardStep>
+              {stepProps => <DaysOfWeekStep {...stepProps} />}
+            </WizardStep>
+          )}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx:109]: I don't think we should actually remove these steps, but use the default values to "jump in" at the desired point. The Wizard accepts an `initialStep` property specifically for this purpose. 

The issue with removing the steps is: 

- Their rows from `SummaryStep` haven't been removed, so these links will currently be broken
- Some steps have bespoke behaviour that depends on other steps, (for example the steps either side of the DayOfWeek DaysOfWeek step jump 2 / -2 depending on whether the session is single or repeating)

Comment (Dave-O-NHS) [reply src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx:109]: Thanks, will update that.

src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx
```diff
@@ -50,9 +53,19 @@ const AvailabilityTemplateWizard = ({ site }: Props) => {
...
+            month: startDate.month() + 1,
+            day: startDate.date(),
+          }
+        : undefined,
     },
   });
   const router = useRouter();
+  const dateIsNotProvided = () => {
+    return typeof date !== 'string';
+  };
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx:68]: It's just a stylistic thing, but personally I actually think this function is less readable than just using the ternary. 
I think `date ? foo : bar` is clearer than `dateIsNotProvided() ? bar : foo`, but up to you :)

src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx
```diff
@@ -62,15 +75,22 @@ const AvailabilityTemplateWizard = ({ site }: Props) => {
...
+    return dateIsNotProvided() ? 1 : 4;
   };
 
   return (
     <FormProvider {...methods}>
       <form onSubmit={methods.handleSubmit(submitForm)}>
         <Wizard
           id="create-availability-wizard"
-          initialStep={1}
+          initialStep={getInitialStep()}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx:93]: Is there any reason `getInitialStep()` is a function? It's called immediately after being declared, so it may as well be declared inline unless I'm missing something? i.e. 
`initialStep={date ? 4 : 1}`

src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx
Comment (Dave-O-NHS) [reply src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx]: Updated, thanks.

src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx
Comment (Dave-O-NHS) [reply src/client/src/app/site/[site]/create-availability/wizard/availability-template-wizard.tsx]: I removed the function as it was not needed.

sambiramairelogic (approved):

PR #394, opened 2025-01-03
Ordering daily appointments

Dave-O-NHS:
Ordering daily appointments by time, first name and last name.

### DIFF
--- a/src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx
+++ b/src/client/src/app/site/[site]/view-availability/daily-appointments/daily-appointments-page.tsx
@@ -73,7 +73,21 @@ export const DailyAppointmentsPage = ({
       headers.push('Action');
     }
 
-    const rows = getPagedBookings(bookings).map(booking => {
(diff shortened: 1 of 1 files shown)


sambiramairelogic (changes requested):
Sorry to be a pain, but personally I think we should do this on the backend rather than the frontend. That way, the sorting is only done once as we'll benefit from NextJS' out of the box caching. 

If it's done here, it'll have to sort the data every time the route is rendered. We only benefit from full route caching really. 

If we were going to keep it in the frontend, we should at least move this code into the appointmentsService so it benefits from NextJS' automatic memoization of server actions.

Dave-O-NHS:
> Sorry to be a pain, but personally I think we should do this on the backend rather than the frontend. That way, the sorting is only done once as we'll benefit from NextJS' out of the box caching.
> 
> If it's done here, it'll have to sort the data every time the route is rendered. We only benefit from full route caching really.
> 
> If we were going to keep it in the frontend, we should at least move this code into the appointmentsService so it benefits from NextJS' automatic memoization of server actions.

PR #395, opened 2025-01-03
APPT-146 - Get booking by reference number endpoint update

ste-banks:
Adding an optional site query string to the GetBookingByReferenceNumber endpoint so we can access it from the front end without having the booking:query permission set globally.


tests/Nhs.Appointments.Api.UnitTests/Validators/QueryBookingByReferenceRequestValidatorTests.cs
sambiramairelogic, line 1: Totally optional but it might be nice to add a couple of extra cases here - one for if site is null because now we've disabled nullable types it implicitly can be, and one for when site holds a happy path value

sambiramairelogic (approved):
Left an optional suggestion so approving anyway

Dave-O-NHS (approved):

PR #396, opened 2025-01-03
Ordering daily appointments

Dave-O-NHS:
Added backend ordering of bookings by time, then first name and finally last name


tests/Nhs.Appointments.Core.UnitTests/BookingServiceTests.cs
sambiramairelogic, line 1: This is an optional suggestion, but it might be nice to add another booking here with the same first name as one of the other bookings to cover the `.ThenBy(b => b.AttendeeDetails.LastName);` line

sambiramairelogic (dismissed):
Left an optional suggestion so approving anyway

sambiramairelogic (approved):

PR #397, opened 2025-01-03
APPT-146 - Update to the cancel appointment endpoint - optional query string for siteId

ste-banks:


src/api/Nhs.Appointments.Api/Functions/CancelBookingFunction.cs
vicr1, line 1: Feels to me that we should support site being passed in the header or perhaps a separate cookie so that signed in users can use these endpoints without us having to extend them with additional params. Might be one for future improvement

src/api/Nhs.Appointments.Core/BookingsService.cs
```diff
@@ -77,11 +77,11 @@ public Task<IEnumerable<Booking>> GetBookingByNhsNumber(string nhsNumber)
         }            
     }
 
-    public async Task<BookingCancellationResult> CancelBooking(string bookingReference)
+    public async Task<BookingCancellationResult> CancelBooking(string bookingReference, string siteId)
```
vicr1, line 80: Something feels off about passing a siteId into this method. I would almost prefer to pass in a delegate
CancelBooking(string bookingReference, Func<Booking, bool> canCancel)) so that the caller has a chance to decide rather than this method, but I'm not sure

src/client/src/app/lib/services/appointmentsService.ts
```diff
@@ -393,8 +393,10 @@ export const fetchBooking = async (reference: string, site: string) => {
   return handleBodyResponse(response);
 };
 
-export const cancelAppointment = async (reference: string) => {
-  const response = await appointmentsApi.post(`booking/${reference}/cancel`);
+export const cancelAppointment = async (reference: string, site: string) => {
```
vicr1, line 396: See previous comment about sending site as a header, might be a future extension that the app always sends the current site as a header ('mya-site'). Again this is really a future improvement suggestion

vicr1 (approved):
Added a few future improvement suggestions

PR #405, opened 2025-01-07
APPT-509: Introduce Orphaned appointment status and recalculate when availability changes

sambiramairelogic:
- Introduces the concept of "Orphaned" appointments and extends the `AppointmentStatus` enum
- Creates a method for programmatically re-assessing the status of every appointment for a given day, adopting any orphans if possible and orphaning any scheduled appointments if needed.
- Calls this method after single day availability is created, availability templates are applied, or appointments are cancelled. 
- Adds tests for all the above


vicr1 (approved):

jsed-nhs (approved):
Looks good to me, the only small thing was around (in future dev work) when an Orphaned appointment is updated to 'Booked' by the user that it may need a new status rather than being set back to 'Booked' - so that it isn't picked up by this Recalculate method a second time - and Re-Orphaned.

PR #408, opened 2025-01-08
Appt 480/filtered list of affected (orphaned) appointments

ste-banks:


src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
```diff
@@ -61,6 +62,23 @@ const Page = async ({ params, searchParams }: PageProps) => {
...
+          {orhpanedAppointments.length > 0 ? (
+            <p>
+              {orhpanedAppointments.length} booked appointments are affected.
+              You'll need to manually cancel these appointments.
+            </p>
+          ) : (
+            <p>
+              There are no booked appointments affected by availability changes.
+            </p>
+          )}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:75]: Can we add some tests for this conditional behaviour? And for the filtering?

src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
```diff
@@ -61,6 +62,23 @@ const Page = async ({ params, searchParams }: PageProps) => {
...
+          <DailyAppointmentsPage
+            bookings={orhpanedAppointments}
+            site={site.id}
+            displayAction
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:79]: We don't yet support cancelling Orphaned appointments so we shouldn't be displaying the `Cancel` buttons just yet - I think displayAction should be false for now. 

(Also whilst we're here, if we're already checking for orphanedAppointments length being 0, we could add this into the ternary so we don't bother rendering it at all if we don't need to)

src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
Comment (ste-banks) [reply src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:75]: Yep, it's on my to do list before publishing the PR :)

sambiramairelogic (approved):

PR #413, opened 2025-01-09
Adding FeedbackBanner component

Dave-O-NHS:


src/client/src/app/lib/components/feedback-banner.tsx
```diff
@@ -0,0 +1,24 @@
+import Tag from '@components/nhsuk-frontend/tag';
+import '../../global.css';
+
+const FeedbackBanner = () => {
+  return (
+    <div className="feedback-banner">
+      <Tag text="Feedback" />
+      &nbsp;
```
sambiramairelogic, line 8: I don't think it's likely to cause problems in our site, but in general we shouldn't be using `&nbsp;` for non-breaking spaces in tsx as it can insert unexpected tags (e.g. in some browsers this becomes `<span></span>` which could cause unexpected changes if we had a css rule targetting span elements)
Just `{'  '}` is the more modern way to handle this and how we're doing it elsewhere

src/client/src/app/lib/components/feedback-banner.tsx
```diff
@@ -0,0 +1,22 @@
+import Tag from '@components/nhsuk-frontend/tag';
+import '../../global.css';
```
sambiramairelogic, line 2: I don't think you need this as `global.css` is already imported through `layout.tsx`

src/client/src/app/global.css
```diff
@@ -236,3 +236,10 @@
 .nhsuk-tabs__list-item::before {
   content: none;
 }
+
+.feedback-banner {
+  text-align: center;
+  padding-top: 8px;
+  padding-bottom: 8px;
+  background-color: #fff;
+}
```
sambiramairelogic, line 245: This styling doesn't look quite right to me - on the designs in the ticket the banner tag/text starts at the same breakpoint as the rest of the site content. Because of `text-align: center` it's appearing out of line:
image

sambiramairelogic (approved):

PR #417, opened 2025-01-10
Appt 481/cancel future session endpoint

ste-banks:
1. New endpoint to cancel a future session - future in this instance can mean a few hours in the future, not the next day or further ahead
2. Once the session has been found and cancelled, we get any bookings within it and orphan them.
3. In the case we find two matching sessions, we just grab the first one as discussed previously.


Dave-O-NHS (approved):

sambiramairelogic (changes requested):
As per comments

src/api/Nhs.Appointments.Core/BookingsService.cs
```diff
@@ -188,4 +189,15 @@ public async Task RecalculateAppointmentStatuses(string site, DateOnly day)
...
+    public async Task OrphanAppointments(string site, DateTime from, DateTime until)
+    {
+        var bookingsInSession = (await GetBookings(from, until, site))
+            .Where(b => b.Status is not AppointmentStatus.Cancelled and not AppointmentStatus.Orphaned);
+
+        foreach (var booking in bookingsInSession)
+        {
+            await SetBookingStatus(booking.Reference, AppointmentStatus.Orphaned);
+        }
+    }
```
sambiramairelogic, line 202: I don't think we should need a new method for this, you should just be able to call `RecalculateAppointmentStatuses` and it'll sort everything for you. 

Similarly though, here you're just orphaning everything but we should be re-scheduling appointments if possible. 
e.g. 2 simultaneous sessions are running with 1 capacity each (so 2 capacity per slot total). 1 appointment is booked into the first session, then that first session is cancelled. Rather than orphaning that appointment it should remain scheduled because it can be fulfilled by the 2nd session

tests/Nhs.Appointments.Api.Integration/Nhs.Appointments.Api.Integration.csproj
```diff
@@ -41,4 +41,8 @@
 	  <None Remove="Nhs.Appointments.Api.Integration\**" />
 	</ItemGroup>
 
+	<ItemGroup>
+	  <None Remove="Scenarios\CancelSession\CancelSession.feature" />
+	</ItemGroup>
```
sambiramairelogic, line 46: Don't think we need these? I assume they've been auto genned by the ide

tests/Nhs.Appointments.Api.Integration/Scenarios/CancelSession/CancelSession.feature
sambiramairelogic, line 1: Could we add a test for the scenario I listed above? i.e. 2 sessions, 1 appointment, one session cancelled but appointment still booked because of the 2nd session?

sambiramairelogic (dismissed):

sambiramairelogic (dismissed):

sambiramairelogic (approved):

PR #419, opened 2025-01-10
Appt 518 - Audit Project and Middleware

jsed-nhs:
This is the first part of the work with the creation of an Audit project class library, that has the middleware and attributes for marking Azure functions to be audited (into CosmosDB)

- New project file added to soln and dockerFile.
- Middleware registration
- Checks for existence of the "RequiresAuditAttribute" on the Azure function as to whether to handle or not, and uses the provided IRequestInspector provided to pull out the siteId (if relevant)
- Hooked up to AuthenticateFunction, CancelBookingFunction, ApplyAvailabilityTemplateFunction, SetAvailabilityFunction functions only for now. However will need to add future functions (such as Editing/Cancelling availability once that work has been completed)
- Unit tests to verify these functions are handled, and that the correct 'SiteId' data is pulled out from the HttpRequest.
- Moved all the 'IRequestInspector' code into the Core project as it is now not just used for Api.Auth purposes and needs to be reusable by both the Api and Audit project. 
- Have moved the associated tests for these Inspectors into the Core.UnitTests project for symmetry.

Still TODO in a follow up PR:

- Hook up writing the data to CosmosDB and setup the schema/partition for the container.
- Full playwright E2E test to confirm Audit data integration for actions
- Consideration of cookie sign in and how that might affect the AuthenticateFunction endpoint
- Ensuring the fire-and-forget of the audit record is consistent and doesn't run into any threading issues.


src/api/Nhs.Appointments.Api/FunctionConfigurationExtensions.cs
```diff
@@ -119,35 +125,48 @@ private static async Task SetupCosmosDatabase(CosmosClient cosmosClient)
         await database.Database.CreateContainerIfNotExistsAsync(id: "index_data", partitionKeyPath: "/docType");
     }
 
+    public static IServiceCollection AddRequestInspectors(this IServiceCollection services)
```
vicr1, line 128: What was the motivation behind moving this here?

src/api/Nhs.Appointments.Audit/Functions/Middleware.cs
```diff
@@ -0,0 +1,64 @@
...
+        var userContextProvider = context.InstanceServices.GetRequiredService<IUserContextProvider>();
+        var requestSiteInspectorType = requiresAudit.RequestSiteInspector;
+        var userId = userContextProvider.UserPrincipal?.Claims.GetUserEmail();
+        var actionType = context.FunctionDefinition.Name;
+        var siteId = string.Empty;
+        if (context.InstanceServices.GetService(requestSiteInspectorType) is IRequestInspector requestInspector)
+        {
+            var request = await context.GetHttpRequestDataAsync();
+
+            //TODO not expecting multiple site matches in first usages??
```
vicr1, line 41: The get availability endpoint is the only one that has multiple sites, it is typically called by an API user, we might need to give some thought about what we want to do in these cases, but single or default will except if this list contains many so this will be a problem on that endpoint

src/api/Nhs.Appointments.Audit/Functions/Middleware.cs
```diff
@@ -0,0 +1,64 @@
...
+        var typeName = string.Join(".", context.FunctionDefinition.EntryPoint.Split('.')[..^1]);
+        var methodName = context.FunctionDefinition.EntryPoint.Split('.')[^1];
+
+        var type = assembly.GetType(typeName);
+        var methodInfo = type?.GetMethod(methodName);
+
+        return methodInfo?.GetCustomAttribute<RequiresAuditAttribute>();
+    }
+
+    protected virtual async Task RecordAudit(Models.Audit auditToWrite)
```
vicr1, line 60: Instead of making this virtual (for testing) would it be sensible to expect the actual writing of the Audit to be done by a separate class behind and IAuditWriter (or similar) interface. Then this could be passed in via constructor injection and the unit testing could pass in a mock and assert that this class is writing the correct audit records?

tests/Nhs.Appointments.Api.UnitTests/Audit/MiddlewareTests.cs
```diff
@@ -0,0 +1,223 @@
...
+        public override ImmutableArray<FunctionParameter> Parameters { get; }
+        public override string PathToAssembly => pathToAssembly;
+        public override string EntryPoint => entrypoint;
+        public override string Id { get; }
+        public override string Name => name;
+        public override IImmutableDictionary<string, BindingMetadata> InputBindings { get; }
+        public override IImmutableDictionary<string, BindingMetadata> OutputBindings { get; }
+    }
+
+    private class TestHttpRequestData(FunctionContext functionContext) : HttpRequestData(functionContext)
```
vicr1, line 178: Have we got this class repeated in a few places?

src/api/Nhs.Appointments.Audit/Functions/Middleware.cs
```diff
@@ -0,0 +1,64 @@
...
+        if (requiresAudit == null)
+        {
+            await next(context);
+            return;
+        }
+        
+        //TODO make sure not awaiting model being built as part of fire and forget?
+        var audit = await BuildAuditModel(context, requiresAudit);
+
+        _ = RecordAudit(audit);
```
vicr1, line 25: I am not sure it really conveys it's intention correctly, i.e. to run the task without awaiting it. Is there are more specific way of doing this? Maybe Task.Run(() => RecordAudit(audit))

src/api/Nhs.Appointments.Api/FunctionConfigurationExtensions.cs
jsed-nhs: It was in the Auth folder under AddCustomAuthentication, however its now not just for Authentication purposes, and the inspectors are in the Core project now. Thought it made more sense at this level as the Inspectors need adding in for FunctionConfiguration instead of AuthenticationConfiguration.

src/api/Nhs.Appointments.Audit/Functions/Middleware.cs
jsed-nhs: Yes the singleOrDefault was intentional so that it would fail fast if we ever audit a function that has multiple sites. Do we get ahead of this and make the 'site' audit property a collection so that we store all? Only thought was it would then not be partitionable maybe...?

src/api/Nhs.Appointments.Audit/Functions/Middleware.cs
jsed-nhs: Yes this will likely be the direction I take it with a follow up PR, haven't thought too hard about the RecordAudit stuff for now because I knew their would be more significant code change to hook it all up. Will refactor the unit tests in a future PR to assert in the way you've described.

src/api/Nhs.Appointments.Audit/Functions/Middleware.cs
jsed-nhs: Yes I think I will do it this way, wanted to save some of the cognitive load around the fire-and-forget writing of the audit to the next PR.

tests/Nhs.Appointments.Api.UnitTests/Audit/MiddlewareTests.cs
jsed-nhs: Yes we seem to, will see if I can reduce it down to 1 to serve all test purposes

vicr1 (approved):

PR #421, opened 2025-01-10
Appt 537/booking data indexing

kicr1:
- Add opt out index on booking_data container 
- Remove spatial index on core_data as no longer needed


infrastructure/resources/database.tf
```diff
@@ -42,6 +42,57 @@ resource "azurerm_cosmosdb_sql_container" "nbs_mya_booking_container" {
...
+      path = "/duration/?"
+    }
+    excluded_path {
+      path = "/service/?"
+    }
+    excluded_path {
+      path = "/status/?"
+    }
+    excluded_path {
+      path = "/statusUpdated/?"
```
vicr1, line 81: statusUpdated is used by daily booking extracts so would make sense to index

vicr1 (changes requested):

PR #428, opened 2025-01-14
Fix for server actions working behind a proxy

vicr1:


sambiramairelogic (changes requested):
lgtm, just a minor thing but since adding the allowedOrigins property to the next config we get this warning on build if it isn't supplied: 
image
Could we add it to the .env or provide a default to avoid triggering this when running locally?

kicr1:
@sambiramairelogic The Allowed origin is no longer needed, so I have removed this from the config and the terraform.

sambiramairelogic (approved):

PR #433, opened 2025-01-15
Creating admin user role

Dave-O-NHS:


tests/Nhs.Appointments.Core.UnitTests/SiteServiceTests.cs
```diff
@@ -522,4 +522,36 @@ public async Task GetSiteByIdAsync_ReturnsDefault_WhenSiteIsNotFound(string scop
...
+                    IntegratedCareBoard: "ICB1",
+                    Location: new Location(Type: "Point", Coordinates: [0.05, 50.0]),
+                    AttributeValues: new List<AttributeValue>() {new (Id: "accessibility/access_need_1", Value: "false")})
+        };
+
+        _siteStore.Setup(x => x.GetAllSites()).ReturnsAsync(sites);
+
+        var result = await _sut.GetAllSites();
+
+        result.Count().Should().Be(sites.Count());
```
sambiramairelogic, line 555: I think we should hard code this to 2 rather than using `sites.Count()`. At the minute if someone mistakenly changed the `sites` variable to an empty list this test would still pass but would no longer be testing what it needs to

sambiramairelogic (changes requested):
I think we need a bit more testing on this - we're not currently testing the `GetUserPermissionsFunction` for example which is what we actually use more so than the `GetUserProfileFunction`.  
I'd also quite like us to add some Playwright tests to prove this works on the frontend

src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
```diff
@@ -37,14 +36,27 @@ protected override async Task<ApiResult<UserProfile>> HandleRequest(EmptyRequest
...
-                var siteInfo = await siteService.GetSiteByIdAsync(site);
-                if(siteInfo != null)
-                    siteInfoList.Add(new UserProfileSite(siteInfo.Id, siteInfo.Name, siteInfo.Address));
+                var allSites = await siteService.GetAllSites();
+
+                foreach (var site in allSites)
+                {
+                    siteInfoList.Add(new UserProfileSite(site.Id, site.Name, site.Address));
+                }
+            }
```
sambiramairelogic, line 49: This might be one to get other dev's opinions on, but actually I'm slightly concerned about returning all the sites here. It'll be fine in beta when there are only 47 sites, but there could be up to 7000 when we go live. 
Because we fetch the User Profile as part of checking the frontend is still logged in, we request it on every page and never cache it, so we'd be fetching 7000 sites on every single day page load. 

Maybe we just add a single "allSites" property or something along those lines instead. @vicr1 @kicr1 would appreciate your thoughts?

kicr1: Get all sites is a very expensive query. If done infrequently then it is likely ok but if done often you will need to look to cache the data I think. 

The site search operation is called very frequently and for this we call allSites so we now cache the site data for 5 or 10minutes(can't remember the exact timing). So that maybe something to look at.

src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
```diff
@@ -54,5 +66,9 @@ protected override Task<IEnumerable<ErrorMessageResponseItem>> ValidateRequest(E
         {
             return Task.FromResult(Enumerable.Empty<ErrorMessageResponseItem>());
         }
+
+        private bool IsAdminUser(User user) {
+            return user.RoleAssignments.Any(ra => ra.Role == "system:admin-user");
```
vicr1, line 71: This approach introduces a "special case" that is not obvious from the way a standard RBAC system works. It is prone to failure if the role name changes but also is not obvious that other abilities are granted to users with this role. Introducing special cases like this often becomes very messy and means that code becomes riddled with if/else statements that are difficult to understand the purpose of

src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
vicr1: Agree that this approach is not optimal.

src/api/Nhs.Appointments.Api/Functions/GetUserProfileFunction.cs
Dave-O-NHS: Thank you for pointing this out. I completely agree that the current approach has limitations, and ideally, we would revisit it to improve the implementation. Given the extremely tight deadline, we've agreed to move forward with this solution for now to meet the immediate requirements. I appreciate you raising this here so that it's documented.

sambiramairelogic:
As mentioned in @kicr1 's recent infrastructure demo, since we're creating a new function in this PR it'll need adding to various terraform files or it'll end up being ran by multiple services. I'm assuming we should disable it in `high_load_functions`, `servicebus_functions`, and `timer_functions` but leave it enabled in `http_functions`?

sambiramairelogic:
Just looking at this AC, please correct me if I'm wrong but I believe at the minute if a user had the admin role with global scope but also any other role at a specific site, there'd still be returned in the manage users screen. Other users with the site config permission would then be able to edit / remove the user. 

I wonder if this needs taking back to Product - if a user at a site also has the admin role what should the behaviour be on the manage users screens?
image

tests/Nhs.Appointments.Core.UnitTests/SiteServiceTests.cs
```diff
@@ -634,4 +634,76 @@ public async Task GetSiteByIdAsync_ReturnsDefault_WhenSiteIsNotFound(string scop
...
+                    Name: "Site 2",
+                    Address: "2 Park Row",
+                    PhoneNumber: "0113 1111111",
+                    OdsCode: "odsCode2",
+                    Region: "R1",
+                    IntegratedCareBoard: "ICB1",
+                    Location: new Location(Type: "Point", Coordinates: [0.05, 50.0]),
+                    AttributeValues: new List<AttributeValue>() {new (Id: "accessibility/access_need_1", Value: "false")})
+        };
+        object? outSites = null;
```
sambiramairelogic, line 699: This will generate build warnings because we've disabled Nullable types. I also notice there are 3 other uses already in this file that must've snuck in from a previous PR, would you mind just resolving them please? Just needs the `?` knocking off

PR #434, opened 2025-01-15
APPT 558 Refactor notifications to fix issues

vicr1:
Fixes an issue where notifications are enqueued even though they do not contain valid contact details and are therefore undeliverable, this will reduce the number of messages enqueued and function run. Also refactored the code to remove repeated classes and improved logging of notification sending. This logging code also stops exceptions being thrown by the notification function. In these cases the messages are undeliverable, throwing an exception leaves the message on the queue and forces a retry which is pointless as the messages can never be delivered.


src/api/Nhs.Appointments.Api/Notifications/BookingNotifier.cs
```diff
@@ -1,19 +1,46 @@
...
+        {
+            logger.LogWarning($"Unable to send notification as event type is unknown");
+            return;
+        }
 
+        if (ValidateDestination(notificationType, destination) == false)
+        {            
+            logger.LogWarning($"Unable to send notification for {eventType}:{bookingRef} to {notificationType}:{destination} because the destination is invalid");
+            return;
+        }
```
sambiramairelogic, line 30: Happy to leave this as it is, but I can't help notice this pattern occurs a lot and I wonder if it might be cleaner to introduce an `IValidator<PatientBookingNotificationEventBase>` (or similar, just whatever the root Notification object is). I wonder if you could then abstract away the "Unknown notification type" logic? 

(though I also notice each method takes a variety of parameters so maybe it wouldn't be as clean as I'm imagining - it's just a thought)

sambiramairelogic (approved):

PR #441, opened 2025-01-18
Appt 574 - Change Site data structure. ID as random GUID, ODSCode as new field.

jsed-nhs:
_Decision was made that the Site model needs updating to facilitate the non-negotiable requirement of editable ODS codes for sites within MYA (they are currently the single internal identifier within the system). Decision made to update the model now before the 'bad' structure makes it any further as edits to ODS need to be possible without having large downstream effects._

**Dev Changes**

- New Id field provided to CSVProcessor should be a randomly generated GUID string for each site. This can be maintained and used by operations onboarding to also apply bulk users to this controlled GUID.
- Update CSVProcessor to move ODSCode onto new field. 
- Add ODSCode field onto SiteDocument CosmosDB document and Site DTO.
- Integration tests updated with new format of ID (for dev consistency) and with new ODS code field.
- Local CosmosDB seeder data updated to match the new ID format (for dev consistency)

- Front end change to use the new OdsCode field for displaying the ODS value in the site details
- Front end tests updated with ODS code change
- Front end tests updated with new ID format (for dev consistency)

**KNOCK-ON EFFECTS**

- The CSVProcessor change impacts the importing data process. The site ID will be new and unique on every run - so huge care needs to be impressed upon operations as possible Site duplicates could leak through with this design change, as no current validation against site name/postcode etc.
- Each environment will need tearing down and re-seeding as the data model has changed.
- The same ODS Codes are now permitted on multiple sites. Operations don't need to increment the ODSCode for satellite sites that share the same ODS Code.
- The URL in MYA uses the new ID (guid) which is less user-friendly, is this a concern for now or tech debt to improve at a later time?
- Questions about the overall data model for the future. Do we need an organisation level document for a relational data? Future tech debt?
- How does this impact the daily export process?
- Reporting - Editing ODS code information means all existing bookings/availability/users are against that new ODS code and the old one is lost (no Site history in the DB). How is this handled in the bigger picture?

### DIFF
--- a/data/CsvDataTool/CsvProcessor.cs
+++ b/data/CsvDataTool/CsvProcessor.cs
@@ -1,11 +1,13 @@
 using System.Globalization;
 using CsvHelper;
 using CsvHelper.Configuration;
-using Nhs.Appointments.Persistance.Models;
 
 namespace CsvDataTool;
 
-public class CsvProcessor<TDocument, TMap>(Func<TDocument, Task> processRow, Func<TDocument, string> getItemName)
+public class CsvProcessor<TDocument, TMap>(
+    Func<TDocument, Task> processRow,
+    Func<TDocument, string> getItemName,
+    Func<TDocument, TDocument> mutateDocument = null)
     where TMap : ClassMap
 {
     public async Task<IEnumerable<ReportItem>> ProcessFile(TextReader csvReader)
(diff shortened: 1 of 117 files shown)


data/CsvDataTool/CsvProcessor.cs
```diff
@@ -31,13 +34,26 @@ public async Task<IEnumerable<ReportItem>> ProcessFile(TextReader csvReader)
...
+            if (typeof(TDocument) == typeof(SiteDocument))
+            {
+                var siteDocs = imported as List<SiteDocument>;
+                foreach (var siteDoc in siteDocs!)
+                {
+                    siteDoc.Id = $"{Guid.NewGuid()}";
+                }
+
+                imported = siteDocs as List<TDocument>;
+            }
```
sambiramairelogic, line 54: If we want to find a better way of doing this, I have 2 thoughts: 

- The first is to create a new document type which inherits from SiteDocument. Some `CsvSiteDocument` or something, and have that new up its own guid if one wasn't provided, then change the type of the process to `CsvProcessor<CsvSiteDocument...`. This still doesn't feel ideal though, and is a bit surprising.
- My preferred idea is to introduce an optional function to the `CsvProcessor`. i.e. something like `Func<TDocument, TDocument> mutateDocument`, then add declare imported as `var imported = csv.GetRecords().Select(mutateDocument)`

sambiramairelogic: The extra benefit of that 2nd idea is that it moves the logic out of `CsvProcessor` and into `SiteDataImportHandler`, which is arguably more where it should live

jsed-nhs: Thats fine I'll do the mutate idea

sambiramairelogic (approved):

PR #444, opened 2025-01-20
APPT-559 - spread functions over multiple app service plans

kicr1:
- Spread functions over 4 different app service plans.
- Functions split by type - http, timer, servicebus.
- Additional app service added for high load functions which hosts only availability at this time
- Add job to deploy and swap slots (when in staging) concurrently and not sequentially


infrastructure/environments/stag/main.tf
```diff
@@ -41,6 +42,8 @@ module "api" {
   splunk_hec_token                               = var.SPLUNK_HEC_TOKEN
   splunk_host_url                                = var.SPLUNK_HOST_URL
   autoscale_notification_email_address           = var.AUTOSCALE_NOTIFICATION_EMAIL_ADDRESS
+  disable_query_availability_function            = true
+  create_high_load_function_app                  = true
```
vicr1, line 46: These only ever seem to be used in tandem, do we need two values or would one suffice?

vicr1 (approved):

PR #447, opened 2025-01-21
APPT-520 - Small permission based UI update

ste-banks:
Only displaying the cancel action on the daily appointments page if a user has permission to cancel them.

### DIFF
--- a/src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
+++ b/src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
@@ -1,5 +1,9 @@
 import NhsPage from '@components/nhs-page';
-import { assertPermission, fetchSite } from '@services/appointmentsService';
+import {
+  assertPermission,
+  fetchPermissions,
+  fetchSite,
+} from '@services/appointmentsService';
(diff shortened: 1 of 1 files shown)


src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx
```diff
@@ -71,7 +79,7 @@ const Page = async ({ params, searchParams }: PageProps) => {
           <DailyAppointmentsPage
             bookings={orphanedAppointments}
             site={site.id}
-            displayAction
+            displayAction={canCancelBookings}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:82]: Minor thing but I've just been solving the same issue on the day view cards - if we remove the element from the row array the table component won't render the wrapping `<td></td>` element, causing the following bug:
image
Could we just amend the following to push the empty string in this case?
image

Comment (ste-banks) [reply src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:82]: For that view it might be thee case but this one is all or nothing and that bug doesn't happen 
image

Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/view-availability/daily-appointments/page.tsx:82]: Ah, my apologies then!

sambiramairelogic (approved):

PR #454, opened 2025-01-23
Appt 565/okta integration

MikeHDigital:
This change adds Okta authentication as an option for user login.

There is a test user set up in the okta organisation, credentials for that will be available in confluence to facilitate testing of this change, however this will need actioning in whatever new development and test organisations that will be set up. Details around the accounts and organisation setup for the deployment to environments are still to be determined.

NHS Mail should continue to work but as yet only been able to test with the dummy oidc locally.

Significant change in this PR would be to the shape of the configuration elements around auth. As there are now multiple auth providers that need their own validation parameters and urls settting up I have altered the configuration structure to facilitate this.

Obviously that will be a breaking change if the configuration for the environments isn't amended before releasing this change set. The intention would be to add this new configuration shape to the environments in addition to the existing configuration before this is released, and then the old configuration shape can be removed once this release has stabilised.


src/api/Nhs.Appointments.Api/Auth/BearerTokenRequestAuthenticator.cs
```diff
@@ -21,25 +27,36 @@ public BearerTokenRequestAuthenticator(ISecurityTokenValidator validator, IJwksR
...
         };
-    
+        
         try
         {
             return _validator.ValidateToken(encodedToken, tokenValidationParams, out var token);
         }
         catch (SecurityTokenValidationException)
         {
-            ClaimsIdentity unauthenticated = new ClaimsIdentity();
```
vicr1, line 41: I think the expanded code helps the reader understand that the returned ClaimsPrincipal is unauthenticated

vicr1 (changes requested):
Additional work is needed to the deployment pipeline to make sure that configuration for the deployed application is correct. As it stands this will not work if deployed as the env vars will not be setup correctly.

src/api/Nhs.Appointments.Api/Auth/BearerTokenRequestAuthenticator.cs
MikeHDigital: I tend to agree, I think that was a bit of overly happy resharper cleanup

vicr1 (approved):

vicr1 (approved):

PR #456, opened 2025-01-23
Appt 491 mya user managent tests

sarithalakkyreddy:
Added user management auto tests


data/CosmosDbSeeder/items/local/core_data/zzz_test_user_8.json
```diff
@@ -0,0 +1,15 @@
...
+    {
+      "role": "canned:appointment-manager",
+      "scope": "site:5914b64a-66bb-4ee2-ab8a-94958c1fdfcb"
+    },
+    {
+      "role": "canned:appointment-manager",
+      "scope": "site:6877d86e-c2df-4def-8508-e1eccf0ea6be"
+    }
+]
+}
```
sambiramairelogic, line 15: Can I recommend you configure VS Code to auto-format files using Prettier so we don't get misaligned files like this? 

If you open the workspace it's declared as a recommended extension so VS Code should prompt you to install it. Failing that you can install it manually through the extensions tab then set it as your default formatter (again if you open the workspace it'll set this for you). It'll then automatically format files whenever you save them so you never have to worry about this again :) 
image

data/CosmosDbSeeder/items/local/core_data/zzz_test_user_7.json
sambiramairelogic, line 1: We don't actually need this file, I think we created it by accident while resolving the merge. It already exists in this folder.

Confusingly, Dave has called this `admin_test_user.json` rather than `zzz_test_uer_7.json` (but it's still using test user 7's username/password in the playwright config). You can just delete this file. 

Optionally if you want to rename admin_test_user.json to zzz_test_uer_7.json feel free

src/client/testing/fixtures.ts
```diff
@@ -13,3 +13,10 @@ export const test = baseTest.extend<object, { newUserName: string }>({
     { scope: 'worker' },
   ],
 });
+
+export const nonNhsEmailId = `int-test-user-${Math.floor(Math.random() * 10000)}@gmail.com`;
```
sambiramairelogic, line 17: You don't need to create this yourself, we already have it as a method on the base test class. 
image
You can use it in a test like so: `test('test name', async ({ newUserName }) => {...`

src/client/testing/page-objects/manage-users/users-page.ts
```diff
@@ -0,0 +1,106 @@
...
+      name: 'Email',
+    });
+    this.manageColumn = page.getByRole('columnheader', {
+      name: 'Manage',
+    });
+    this.assignStaffRolesLink = page.getByRole('link', {
+      name: 'Assign Staff Roles',
+    });
+    this.removeUserSuccessBanner = page.getByText(
+      'You have successfully removed zzz_test_user_2@nhs.net from the current site.',
```
sambiramairelogic, line 26: We should probably avoid hardcoding this in the page object as ideally methods in here should be re-usable. 

We could make it a function which takes the email as a parameter, but to be honest I think this would be better off just asserted in the test itself

src/client/testing/page-objects/oauth.ts
```diff
@@ -22,4 +22,10 @@ export default class OAuthLoginPage extends RootPage {
 
     await this.page.waitForURL(env.BASE_URL);
   }
+
+  async signInWithRequiredUser(userName: string, password: string) {
+    await this.page.getByLabel('Username').fill(userName);
+    await this.page.getByLabel('Password').fill(password);
+    await this.page.getByLabel('Password').press('Enter');
+  }
```
sambiramairelogic, line 30: We already have this method in this file just above this one

src/client/testing/page-objects/site.ts
```diff
@@ -19,5 +21,44 @@ export default class SitePage extends RootPage {
...
   }
+
+  //   async veriyTileNotViisible(){
+  //     // await expect(
+  //     //    sitePage.viewAvailabilityAndManageAppointmentsCard,
+  //     //  ).toBeVisible();
+  //     //  await expect(sitePage.createAvailabilityCard).not.toBeVisible();
+  //     //  await expect(sitePage.userManagementCard).not.toBeVisible();
+  //     //  await expect(sitePage.siteManagementCard).toBeVisible();
+  //  };
```
sambiramairelogic, line 63: commented code

data/CosmosDbSeeder/items/local/core_data/zzz_test_user_7.json
sarithalakkyreddy: Deleted zzz_test_uer_7.json extra file and renamed admin_test_user.json to zzz_test_uer_7.json

data/CosmosDbSeeder/items/local/core_data/zzz_test_user_8.json
sarithalakkyreddy: Done

src/client/testing/fixtures.ts
sarithalakkyreddy: I have define this one to generate non NHS email Id.

src/client/testing/page-objects/manage-users/users-page.ts
sarithalakkyreddy: Removed that locator as we already defined another function where we handling email dynamically. 
eg: `You have successfully removed ${userName} from the current site.`,

src/client/testing/page-objects/oauth.ts
sarithalakkyreddy: Designed second function to sign in with any user, earlier function was hardcoded with one user which is Test user1 going forward will make changes in code to use second function and will remove first one.

src/client/testing/page-objects/site.ts
sarithalakkyreddy: Removed the commented code

src/client/testing/fixtures.ts
sambiramairelogic: Fair enough, but I still think we should be consistent and define it the same as the nhs username has been. The benefits of this are:

- We don't need to implement a random number generator ourselves
- We can rely on it being the same on each reference, but different between each test (the const will only be different each time it's imported)
- It's consistent with the way we're already doing this, rather than introducing a new pattern

image

src/client/testing/page-objects/oauth.ts
sambiramairelogic: That's incorrect, the existing function lets you sign in with any user you want. `ser: { username: string; password: string } = env.TEST_USERS.testUser1` means that the parameters `username` and `password` are optional. If not provided, they will use test user 1 by default, but if they are provided they'll use the provided values instead. 

There is an example of this in  `user-management-permissions.spec.ts` => `await oAuthPage.signIn(TEST_USERS.testUser2);`

azure-pipelines-bulk-data.yml
sambiramairelogic, line 1: Did you intend to make these changes? Looks like this might have been a rebase mistake?

data/CosmosDbSeeder/items/dev/core_data/notification_config.json
sambiramairelogic, line 1: Same here?

src/client/testing/fixtures.ts
sarithalakkyreddy: Fixed

azure-pipelines-bulk-data.yml
sarithalakkyreddy: I haven't changed anything there

src/api/Nhs.Appointments.Core/BookingsService.cs
```diff
@@ -182,7 +182,7 @@ public async Task RecalculateAppointmentStatuses(string site, DateOnly day)
                 continue;
             }
 
-            if (booking.Status is AppointmentStatus.Booked or AppointmentStatus.Provisional)
+            if (booking.Status is AppointmentStatus.Booked)
```
jsed-nhs, line 185: Is this change intentional or a rebase error?

tests/Nhs.Appointments.Core.UnitTests/BookingsServiceTests.cs
```diff
@@ -518,7 +518,7 @@ public async Task RecalculateAppointmentStatuses_OrphansLiveAppointmentsIfTheyCa
                         From = new DateTime(2025, 01, 01, 9, 10, 0),
                         Reference = "2",
                         AttendeeDetails = new AttendeeDetails { FirstName = "Alexander", LastName = "Cooper" },
-                        Status = AppointmentStatus.Provisional,
+                        Status = AppointmentStatus.Booked,
```
jsed-nhs, line 521: Is this change intentional or rebase error? surely this PR shouldn't be changing any Booking files?

src/client/testing/page-objects/oauth.ts
sarithalakkyreddy: Done

tests/Nhs.Appointments.Core.UnitTests/BookingsServiceTests.cs
sarithalakkyreddy: Its a rebase error

src/api/Nhs.Appointments.Core/BookingsService.cs
sarithalakkyreddy: Its a rebase error

azure-pipelines-bulk-data.yml
sambiramairelogic, line 1: Looks like we still have some rebase errors (this file for example). Can you go through the files changed and let me know when all the changes are ones you intend to make please?

data/CosmosDbSeeder/items/dev/core_data/notification_config.json
sarithalakkyreddy, line 1: Rebase error

src/api/Nhs.Appointments.Api/Notifications/template-bookingcancelled-rsv-sms.txt
sarithalakkyreddy, line 1: Rebase error

src/api/Nhs.Appointments.Core/BookingsService.cs
sarithalakkyreddy, line 1: Rebase error

src/api/Nhs.Appointments.Api/Notifications/template-userroleschanged.txt
sarithalakkyreddy, line 1: Rebase error

azure-pipelines-bulk-data.yml
sarithalakkyreddy: Below are files which got updated after Rebase
data/CosmosDbSeeder/items/dev/core_data/notification_config.json
 src/api/Nhs.Appointments.Api/Notifications/template-bookingcancelled-rsv-email.txt
src/api/Nhs.Appointments.Api/Notifications/template-bookingcancelled-rsv-sms.txt
src/api/Nhs.Appointments.Api/Notifications/template-userroleschanged.txt
src/api/Nhs.Appointments.Core/BookingsService.cs
tests/Nhs.Appointments.Core.UnitTests/BookingsServiceTests.cs

PR #459, opened 2025-01-24
Appt 385 - Edit Site Details (Core)

jsed-nhs:
Edit Site Details Page and Form
Split Site details into Core and Admin (ODS, Region,..) cards in the Details page.
Gherkin integration tests happy path and invalid data paths
Jest tests for form submission parsing (address over multiple lines in textarea, strip spaces out of phoneNumber field)
E2E Playwright test for Saving details, or cancelling halfway through.
Refactor Information and Access details tests to cleanup data after themselves so can be run in succession.
Refactor Jest summary-list file to have more targeted aria-label for the summary list. This enabled jest tests to be more targeted about the data in each listitem, and makes it more readable.

**TODO in follow up PRs:**
Found concurrency issue with information and access tests in Playwright running in parallel - will need API service change to retry when document has changed.
Admin details edit page (pretty similar to this work but targeted at the admin data instead)


data/CosmosDbSeeder/items/local/core_data/site_ABC01.json
```diff
@@ -3,7 +3,7 @@
   "docType": "site",
   "name": "Robin Lane Medical Centre",
   "address": "Pudsey, Leeds, LS28 7BR",
-  "phoneNumber": "0113 1111111",
+  "phoneNumber": "01131111111",
```
jsed-nhs, line 6: Only changing these so Playwright tests work round-trip, spaces is permitted on bulk-import, but then will be resaved without any if ever edited. (Confirmed with BA/UI)

sambiramairelogic: What's the reason for this decision? Seems needlessly user-unfriendly to me

src/client/src/app/site/[site]/details/edit-details/edit-details-form.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+jest.mock('@services/appointmentsService');
+
+jest.mock('next/navigation');
+const mockUseRouter = useRouter as jest.Mock;
+const mockReplace = jest.fn();
+
+const mockSaveSiteDetails = jest.spyOn(appointmentsService, 'saveSiteDetails');
+
+let user: UserEvent;
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/details/edit-details/edit-details-form.test.tsx:17]: This use of `UserEvent` is now an anti-pattern (as of about a year ago). They now want you to deconstruct the `user` object from the `render` call itself. We do this in `src/app/testing/render` which you'll see imported in other tests. 
Correct usage is 
`const { user } = render(<Foobar/>);`
`await user.click(.....);`

Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/details/edit-details/edit-details-form.test.tsx:17]: Docs if curious https://testing-library.com/docs/user-event/setup/

src/api/Nhs.Appointments.Persistance/SiteStore.cs
```diff
@@ -50,6 +50,37 @@ public async Task<OperationResult> UpdateSiteAttributes(string siteId, string sc
...
+    public async Task<OperationResult> UpdateSiteDetails(string siteId, string name, string address, string phoneNumber, string latitude, string longitude)
+    {
+        //TODO move away parsing?
+        var latitideParseResult = decimal.TryParse(latitude, out var latitudeDecimal);
+        var longitudeParseResult = decimal.TryParse(longitude, out var longitudeDecimal);
+
+        if (!latitideParseResult || !longitudeParseResult)
+        {
+            return new OperationResult(false, "The specified lat/long values are not valid decimals.");
+        }
```
sambiramairelogic, line 62: Can't we create an `IValidator<>` for the request these came from and validate them before they hit the function? We should be able to do this check higher up and pass decimals rather than strings down to the service/store

src/api/Nhs.Appointments.Api/Validators/SetSiteDetailsValidator.cs
```diff
@@ -0,0 +1,34 @@
...
+            .Matches(NumbersOnlyRegex).WithMessage("Phone number must contain numbers only")
+            .NotEmpty().WithMessage("Provide a valid phone number");
+        RuleFor(x => x.Latitude)
+            .Matches(DecimalRegex).WithMessage("Latitude must be a decimal number")
+            .NotEmpty()
+            .WithMessage("Provide a valid latitude");
+        RuleFor(x => x.Longitude)
+            .Matches(DecimalRegex).WithMessage("Longitude must be a decimal number")
+            .NotEmpty()
+            .WithMessage("Provide a valid longitude");
```
sambiramairelogic, line 32: Any reason we've gone with Regex over a TryParse?

src/client/testing/tests/change-site-details-tests/edit-details.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  expect(editDetailsPage.longitudeInput).toHaveValue('53.742');
+  expect(editDetailsPage.phoneNumberInput).toHaveValue('01189998819991197253');
+
+  //reset data at end of test
+  await editDetailsPage.addressInput.fill('Pudsey,\nLeeds,\nLS28 7LD');
+  await editDetailsPage.latitudeInput.fill('-1.66382134');
+  await editDetailsPage.longitudeInput.fill('53.79628754');
+  await editDetailsPage.phoneNumberInput.fill('01132222222');
+
+  await editDetailsPage.saveAndContinueButton.click();
```
sambiramairelogic, line 132: If this test is running in serial, why is this necessary? 
If it is necessary, I'd argue it should be in a `test.afterAll()` method rather than in the body of the test

src/client/testing/page-objects/change-site-details-pages/site-details.ts
```diff
@@ -5,25 +5,48 @@ import RootPage from '../root';
...
+  readonly addressLabel = 'Address';
+  readonly latitudeLabel = 'Latitude';
+  readonly longitudeLabel = 'Longitude';
+  readonly phoneNumberLabel = 'Phone Number';
+
+  readonly defaultSiteName = 'Church Lane Pharmacy';
+  readonly defaultAddress = 'Pudsey, Leeds, LS28 7LD';
+  readonly defaultLatitude = '-1.66382134';
+  readonly defaultLongitude = '53.79628754';
+  readonly defaultPhoneNumber = '01132222222';
```
sambiramairelogic, line 26: A pattern I've been musing on for awhile is to expand the constructors of some of the page object to accept relevant data. In this case we could abstract away the sites into config (like the users are currently), then new up the page object with `new SiteDetailsPage(page, sites.siteFoo)`. Sites could be optional and default to this one. Means less hardcoding of mock data

src/api/Nhs.Appointments.Api/Validators/SetSiteDetailsValidator.cs
jsed-nhs: Absolutely no idea! good spot will fix!

src/client/src/app/site/[site]/details/edit-details/edit-details-form.test.tsx
Comment (jsed-nhs) [reply src/client/src/app/site/[site]/details/edit-details/edit-details-form.test.tsx:17]: But this user variable is initialised in the beforeEach using that same render method that you mentioned, which means setup will be called for each one? unless i misunderstand you?

Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/details/edit-details/edit-details-form.test.tsx:17]: Oh yes, you're right, sorry. I just saw the `import { UserEvent }` and 
Hoisting the render up into the beforeEach does feel like a bit of an over-abstraction and will need un-doing if we ever want to pass different arguments in to test different things, but fair enough :)

data/CosmosDbSeeder/items/local/core_data/site_ABC01.json
jsed-nhs: Agreed will push back on this. Initial thought was it might make validation of input easier for MVP without spaces but will have a follow up with them.

jsed-nhs: Have made this change, thanks for the pushback :)

src/client/testing/tests/change-site-details-tests/edit-details.spec.ts
jsed-nhs: Have added to afterAll. I personally like the tests resetting their data so that they can be ran multiple times in succession and still pass :)

src/client/testing/page-objects/change-site-details-pages/site-details.ts
jsed-nhs: I kind of see what you mean, but maybe its clearer in your head, so will leave for now and can be refactored later.

sambiramairelogic (approved):

PR #460, opened 2025-01-24
APPT-592: Return Orphaned bookings from QueryBookingByNhsNumber

sambiramairelogic:
`GET api/booking?nhsNUmber=123` derives a list of sites a person has bookings at but filters it by status = `Booked`. It then queries all their bookings at that site and returns them regardless of status. 

This means if a citizen only has Orphaned bookings at a given site, these will not be included in the response. They need to be so NBS can detect when a user has a live booking.


vicr1 (approved):

z-williams:
We're arranging a meeting for next week to understand what should be done with orphaned bookings from an NBS public perspective. Can this be put on hold until then?

vicr1:
> We're arranging a meeting for next week to understand what should be done with orphaned bookings from an NBS public perspective. Can this be put on hold until then?

NBS should treat these as booked appointments. The concept of being orphaned should not exist outside of MYA, it is an internal concept.

sambiramairelogic:
This change will not be made - we will re-implement the concept of Orphaned appointments in another way (in https://nhsd-jira.digital.nhs.uk/browse/APPT-594)

PR #464, opened 2025-01-27
APPT-XXX: Parallelise data fetches

sambiramairelogic:
Cheeky off-ticket PR to improve data fetching performance on the frontend. 

We're currently doing multiple server actions sequentially which hasn't stung us yet because they're all quite quick, but there's no reason we can't fire off these requests in parallel.

### DIFF
--- a/src/client/src/app/site/[site]/appointment/[reference]/cancel/page.tsx
+++ b/src/client/src/app/site/[site]/appointment/[reference]/cancel/page.tsx
@@ -16,10 +16,12 @@ type PageProps = {
 };
 
 const Page = async ({ params }: PageProps) => {
-  const site = await fetchSite(params.site);
-  await assertPermission(site.id, 'booking:cancel');
+  const [site, booking] = await Promise.all([
+    fetchSite(params.site),
+    fetchBooking(params.reference, params.site),
+    assertPermission(params.site, 'booking:cancel'),
+  ]);
 
-  const booking = await fetchBooking(params.reference, site.id);
   if (!booking || booking.status === 'Cancelled') {
     notFound();
   }
--- a/src/client/src/app/site/[site]/create-availability/wizard/page.tsx
+++ b/src/client/src/app/site/[site]/create-availability/wizard/page.tsx
@@ -1,4 +1,4 @@
-import { fetchSite } from '@services/appointmentsService';
+import { assertPermission, fetchSite } from '@services/appointmentsService';
 import NhsTransactionalPage from '@components/nhs-transactional-page';
 import AvailabilityTemplateWizard from './availability-template-wizard';
 
@@ -12,7 +12,10 @@ type PageProps = {
 };
 
 const Page = async ({ params, searchParams }: PageProps) => {
-  const site = await fetchSite(params.site);
+  const [site] = await Promise.all([
+    fetchSite(params.site),
+    assertPermission(params.site, 'availability:setup'),
+  ]);
 
   return (
     <NhsTransactionalPage originPage="create-availability-wizard">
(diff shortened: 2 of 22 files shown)


src/client/src/app/site/[site]/appointment/[reference]/cancel/page.tsx
```diff
@@ -16,10 +16,12 @@ type PageProps = {
 };
 
 const Page = async ({ params }: PageProps) => {
-  const site = await fetchSite(params.site);
-  await assertPermission(site.id, 'booking:cancel');
+  const [site, booking] = await Promise.all([
+    fetchSite(params.site),
+    fetchBooking(params.reference, params.site),
+    assertPermission(params.site, 'booking:cancel'),
```
Comment (jsed-nhs) [inline src/client/src/app/site/[site]/appointment/[reference]/cancel/page.tsx:22]: Personally feel here that we should maybe await the permission check first, and then can fetch both site and booking needed to construct the page, if successful?

src/client/src/app/site/[site]/create-availability/wizard/page.tsx
```diff
@@ -12,7 +12,10 @@ type PageProps = {
 };
 
 const Page = async ({ params, searchParams }: PageProps) => {
-  const site = await fetchSite(params.site);
+  const [site] = await Promise.all([
+    fetchSite(params.site),
+    assertPermission(params.site, 'availability:setup'),
```
Comment (jsed-nhs) [inline src/client/src/app/site/[site]/create-availability/wizard/page.tsx:17]: Same here...?

src/client/src/app/site/[site]/page.tsx
```diff
@@ -22,11 +22,12 @@ type PageProps = {
...
 const Page = async ({ params }: PageProps) => {
-  const site = await fetchSite(params.site);
-  const wellKnownOdsCodeEntries = await fetchWellKnownOdsCodeEntries();
-  const sitePermissions = await fetchPermissions(params.site);
-
-  await assertPermission(site.id, 'site:view');
+  const [site, wellKnownOdsCodeEntries, sitePermissions] = await Promise.all([
+    fetchSite(params.site),
+    fetchWellKnownOdsCodeEntries(),
+    fetchPermissions(params.site),
```
Comment (jsed-nhs) [inline src/client/src/app/site/[site]/page.tsx:28]: assertPermissions is calling fetchPermissions under the hood, do we really need two calls here? (even if cached?)

src/client/src/app/site/[site]/users/page.tsx
```diff
@@ -16,13 +16,15 @@ type PageProps = {
...
-
-  await assertAnyPermissions(site.id, ['users:view', 'users:view']);
+  const [userProfile, users, rolesResponse, site, permissions] =
+    await Promise.all([
+      fetchUserProfile(),
+      fetchUsers(params.site),
+      fetchRoles(),
+      fetchSite(params.site),
+      fetchPermissions(params.site),
+      assertAnyPermissions(params.site, ['users:view', 'users:view']),
```
Comment (jsed-nhs) [inline src/client/src/app/site/[site]/users/page.tsx:26]: Same here for permissions?

jsed-nhs (changes requested):

src/client/src/app/site/[site]/page.tsx
Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/page.tsx:28]: Most pages only care that a user has a permission, but this page needs to actually enumerate the permissions to drive a UI element. You're right we could knock off `assertPermissions` and replace it with a manual check, but that'd be deviating from the otherwise established pattern. I'd rather keep the pattern and rely on the cache here, but can switch up if you really want

src/client/src/app/site/[site]/appointment/[reference]/cancel/page.tsx
Comment (jsed-nhs) [reply src/client/src/app/site/[site]/appointment/[reference]/cancel/page.tsx:22]: I don't think its risky, thought more just unnecessary calls to the API if its a quick initial check whether the user is permitted to the page, and no need to hit the API twice if know its gonna fail? Conversely, if we have backend secured permission checks, what is the assertPermission achieving - couldn't you remove it and instead handle a 403 response from either of the site/booking responses?

sambiramairelogic:
Nevermind.

PR #465, opened 2025-01-27
APPT-587: Delete provisional bookings when their availability is removed

sambiramairelogic:
Subsequent to the previous PR which prevented Provisional appointments from being orphaned when their availability is removed, we actually want to hard delete them in this scenario. 

I have misgivings about hard deletes, but we're doing this during the nightly sweep so there is precedent, and after all it is only provisional.


jsed-nhs:
Any reason we aren't soft deleting, just to free up data - no current predicted use for this data I assume? Only alternative could be soft delete and then hard delete after X amount of time, you mentioned some "nightly sweep" process?

src/api/Nhs.Appointments.Persistance/BookingCosmosDocumentStore.cs
```diff
@@ -184,10 +184,15 @@ public async Task<IEnumerable<string>> RemoveUnconfirmedProvisionalBookings()
...
-            await bookingStore.DeleteDocument(indexDocument.Reference, indexDocument.Site);
+            await DeleteBooking(indexDocument.Reference, indexDocument.Site);
         }
 
         return indexDocuments.Select(i => i.Reference).ToList();
     }
+
+    public async Task DeleteBooking(string reference, string site)
+    {
+        await indexStore.DeleteDocument(reference, "booking_index");
```
jsed-nhs, line 195: Can we WaitAll/WhenAll for these two? assume no dependency between them?

jsed-nhs (changes requested):
Approved when both async Delete calls run in parallel

src/api/Nhs.Appointments.Persistance/BookingCosmosDocumentStore.cs
sambiramairelogic: Done

jsed-nhs (approved):

PR #467, opened 2025-01-28
APPT-595 - Apply permissions to the manage availability links

ste-banks:
On the week view for availability we have three actions - 'Change', 'Add Session', 'Add availability to this day'. This change removes these links if the user doesn't have the `availability:setup` permission.


jsed-nhs (changes requested):
Any reason you are fetching the permission on the view-week page and then passing down, rather than fetching it in the day-summary alone? Doesn't seem like its used for anything on the view-week page other than drilling down?
Can a test be added to the day-summary-card to verify outcome when permission is false?

ste-banks:
> Any reason you are fetching the permission on the view-week page and then passing down, rather than fetching it in the day-summary alone? Doesn't seem like its used for anything on the view-week page other than drilling down? Can a test be added to the day-summary-card to verify outcome when permission is false?

Because the `day-summary` component is rendered 7 times on this page so we'd end up fetching permissions 7 times instead of 1. And have added a couple of extra tests.

jsed-nhs:
Completely missed the .map around it, fair enough!

jsed-nhs (approved):

PR #472, opened 2025-01-28
Appt 491/user management tests

jsed-nhs:


src/client/testing/page-objects/change-site-details-pages/site-details.ts
```diff
@@ -18,20 +19,38 @@ export default class SiteDetailsPage extends RootPage {
...
+      location: {
+        coordinates: [-1.6610648, 53.795467],
+        type: 'point',
+      },
+      phoneNumber: '0113 1111111',
+      odsCode: 'ABC01',
+      integratedCareBoard: 'Integrated Care Board 1',
+      region: 'Region 1',
+    },
+  ];
```
sambiramairelogic, line 53: We shouldn't be putting hardcoded seed data in our page objects. This will end up getting duplicated the next time another page object needs it or becoming really hard to maintain and increasing the chance of manual errors. 

All data like this should be imported from an external source. We could put it into the config, or create a separate seed data folder within the playwright structure. 

We then only populate the page object with seed data at runtime, either by passing the data in as arguments to methods or as a property within the page object constructor.

src/client/testing/page-objects/change-site-details-pages/site-details.ts
```diff
@@ -131,23 +150,41 @@ export default class SiteDetailsPage extends RootPage {
...
+  }
+
+  async verifySitePage(
+    siteName: 'Church Lane Pharmacy' | 'Robin Lane Medical Centre',
+  ) {
+    const site = this.sites.find(x => x.name == siteName);
+
+    if (site === undefined) {
+      throw new Error();
+    }
```
sambiramairelogic, line 170: This is one of the reasons we shouldn't put seed data in page objects. If our seed data changed or we added extra sites, you'd have to change not only the lines above but also this method too. 
Ideally we only want to have to declare the seed data in one place then we only have to change it once in the future, if that makes sense?

src/client/testing/fixtures.ts
```diff
@@ -4,12 +4,28 @@ process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
...
+      await use(nonNHSuserName);
+    },
+    { scope: 'worker' },
+  ],
 });
+
+export const testuser8_emailId = `zzz_test_user_8@nhs.net`;
+export const testuser9_emailId = `zzz_test_user_9@nhs.net`;
+export const testuser10_emailId = `zzz_test_user_10@nhs.net`;
+export const testuser11_emailId = `zzz_test_user_11@nhs.net`;
```
sambiramairelogic, line 30: I think it'd be clearer to store this information alongside the usernames/passwords. These are currently declared in the .env file and piped through as config.

If we move the sites out into a new seed-data folder as I've suggested below, we could also move the users data out there too. The existing pattern of piping them through from env doesn't really work anymore now the data we need is getting larger

src/client/testing/fixtures.ts
```diff
@@ -4,12 +4,28 @@ process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
...
+export const abc01_id = `5914b64a-66bb-4ee2-ab8a-94958c1fdfcb`;
```
sambiramairelogic, line 31: You then wouldn't need this at all, because it'd come from the seed data itself

src/client/testing/fixtures.ts
```diff
@@ -2,14 +2,80 @@
...
   newUserName: [
     async ({}, use) => {
       const userName = `int-test-user-${test.info().workerIndex}@nhs.net`;
       await use(userName);
     },
     { scope: 'worker' },
   ],
+  externalUserName: [
+    async ({}, use) => {
+      const nonNHSuserName = `external-user-${test.info().workerIndex}@gmail.com`;
```
sambiramairelogic, line 76: Miniscule change so only worth doing if you're making any other change along with it, but camelcase typo in `nonNHSuserName`. I don't know if it should be `nonNHSUserName` or `nonNhsUserName`. The latter is more consistent with what we're doing elsewhere with ODS code etc.

src/client/testing/page-objects/change-site-details-pages/site-details.ts
```diff
@@ -158,17 +163,45 @@ export default class SiteDetailsPage extends RootPage {
...
+    switch (this.siteDetails.region) {
+      case 'R1':
+        expectedRegionDisplayValue = 'Region 1';
+        break;
+      case 'R2':
+        expectedRegionDisplayValue = 'Region 2';
+        break;
+      default:
+        expectedRegionDisplayValue = this.siteDetails.region;
+        break;
```
sambiramairelogic, line 199: This can be a future improvement if you like, but we have already discussed/agreed it would be better to move the `verifyFoo` calls out of the page objects. If we moved these two Region/ICB assertions out into the tests we'd also remove the need to do this mapping at all

jsed-nhs: Yes aware of the tech debt of asserting in the PageObjects, saving for a later time, don't want to do everything at once.

src/client/testing/page-objects/change-site-details-pages/site-details.ts
```diff
@@ -131,23 +126,33 @@ export default class SiteDetailsPage extends RootPage {
...
+    await expect(this.editSiteAttributesButton).toBeVisible();
+  }
+
+  async verifySitePage() {
     await expect(
       this.page.getByRole('heading', { name: `${this.headerMsg}` }),
     ).toBeVisible();
     await expect(
-      this.page.getByRole('heading', { name: `${this.defaultSiteName}` }),
+      this.page.getByRole('heading', { name: `${this.siteDetails?.name}` }),
```
sambiramairelogic, line 144: You don't need the optional chaining operator here

src/client/testing/fixtures.ts
```diff
@@ -2,14 +2,80 @@
...
+const testUsersData: UserSeedData[] = testUsersDataRaw;
+const testSite1Data: SiteWithAttributes = testSite1DataRaw;
+const testSite2Data: SiteWithAttributes = testSite2DataRaw;
+
+export interface UserSeedData {
+  Username: string;
+  Password: string;
+  SubjectId: string;
+}
+
```
sambiramairelogic, line 22: Any reason we've broken with the camelcase convention here? We're using it on all our other types/interfaces

jsed-nhs: Fewer changes, agree though will standardise

PR #479, opened 2025-01-30
Appt 565/okta integration

MikeHDigital:
Adding Okta auth provider

- This updated change defaults the auth provider to the first configured provider if no provider query string is being round-tripped by the auth provider.

This is to enable us to add Okta auth without waiting for changes to the NHS mail application configuration which we can't do as a self-service operation.


src/api/Nhs.Appointments.Api/Functions/AuthenticateFunction.cs
```diff
@@ -21,7 +21,7 @@ public IActionResult Run(
     {
         // here we need to resolve the correct auth provider
         var providerName = req.Query["provider"];
-        var authProvider = authOptions.Value.Providers.Single(p => p.Name == providerName);
+        var authProvider = string.IsNullOrWhiteSpace(providerName) ? authOptions.Value.Providers[0] : authOptions.Value.Providers.Single(p => p.Name == providerName);
```
vicr1, line 24: This is not required, as this is invoked by the front end which should always send the QS

src/api/Nhs.Appointments.Api/Functions/GetAuthTokenFunction.cs
```diff
@@ -31,7 +31,7 @@ public async Task<IActionResult> RunAsync(
     {
         // need to resolve the correct auth config here
         var providerName = req.Query["provider"];
-        var authProvider = authOptions.Value.Providers.Single(p => p.Name == providerName);
+        var authProvider = string.IsNullOrWhiteSpace(providerName) ? authOptions.Value.Providers[0] : authOptions.Value.Providers.Single(p => p.Name == providerName);
```
vicr1, line 34: This is not required as this is invoked by the front end which should always send the QS

vicr1 (approved):

src/api/Nhs.Appointments.Api/Functions/AuthenticateFunction.cs
MikeHDigital: Ahh yeah, this one is the FE function, but this wont really hurt. I'll revert this file anyway

jsed-nhs:
Is the QS change not needed in the other tf files?

jsed-nhs (approved):

PR #484, opened 2025-01-31
APPT-525 - Add site search functionality on the front end. 

ste-banks:
image
image
image

Search bar for filtering sites:
- Requires at least 3 characters to be input before filtering
- Case insensitive
- Will match on ODS _only_ if it matches **exactly**. 
- Exposing the ODS code on the _sites-preview_ endpoint so we can filter on it
- There were no designs on the ticket itself for the search input so I've gone with this by using the existing TextBox component

Out of scope for APPT-525 but the _sites-preview_ endpoint will return all sites a user has permission to without pagination or without limiting the number of sites returned at a given time. It might be worth considering if we need to add this in at some point in the future, either for MVP or post-MVP

### DIFF
--- a/src/client/src/app/lib/components/site-list.tsx
+++ b/src/client/src/app/lib/components/site-list.tsx
@@ -4,6 +4,7 @@ import { Site } from '@types';
 import { Card, TextInput } from '@nhsuk-frontend-components';
 import { sortSitesByName } from '@sorting';
 import { ChangeEvent, useState } from 'react';
+import _debounce from 'lodash.debounce';
 
 type Props = {
   sites: Site[];
@@ -28,13 +29,15 @@ const SiteList = ({ sites }: Props) => {
     }
   };
 
+  const debounceSearchHandler = _debounce(handleInputChange, 300);
+
   return (
     <Card title="Choose a site">
       <TextInput
         id="site-search"
         aria-label="site-search"
         placeholder="Search"
-        onChange={handleInputChange}
+        onChange={debounceSearchHandler}
       ></TextInput>
       <ul className="nhsuk-list nhsuk-list--border">
         {filteredSites.map(s => (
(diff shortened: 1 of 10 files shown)


src/client/src/app/lib/components/site-list.tsx
```diff
@@ -1,19 +1,43 @@
...
 type Props = {
   sites: Site[];
 };
 
 const SiteList = ({ sites }: Props) => {
   const sortedSites = sites.toSorted(sortSitesByName);
+  const [filteredSites, setFilteredSites] = useState(sortedSites);
+
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const searchInput = e.target.value.toLowerCase();
```
jsed-nhs, line 17: Can we debounce this filter input change call? I get it isn't firing off API calls, but with all 7k+ sites in Prod the front end filtering could be slow if type 10 characters in quick succession. Can either use lodash or custom react debounce hook, with a 300-500ms debounce timer?

src/client/src/app/lib/components/site-list.tsx
```diff
@@ -1,19 +1,43 @@
...
+            s.name.toLowerCase().includes(searchInput) ||
+            s.odsCode.toLowerCase() === searchInput,
+        ),
+      );
+    } else {
+      setFilteredSites(sortedSites);
+    }
+  };
 
   return (
```
jsed-nhs, line 31: Can we add a simple playwright test in to confirm functionality?

src/client/src/app/lib/components/site-list.tsx
```diff
@@ -1,19 +1,43 @@
...
     <Card title="Choose a site">
+      <TextInput
+        id="site-search"
+        aria-label="site-search"
+        placeholder="Search"
+        onChange={handleInputChange}
+      ></TextInput>
```
jsed-nhs, line 38: Maybe a little bit of padding/margin so the search isn't right next to the top listitem

jsed-nhs (changes requested):

jsed-nhs (approved):
Would prefer the debounce be a hook, but okay as it is for now.

PR #486, opened 2025-01-31
Appt 502 - Edit site reference details, front end validation for all site details

jsed-nhs:
Edit Site Reference details page and form, locked behind the 'system:admin-user' permission
Added front end regex validation for the core site details and the reference details
Validate against empty textboxes for both details
Gherkin, jest and playwright tests for edit reference details
Auditing the 3 set site details functions
Added a new local seed site that doesn't have a well-known ICB/Region codes
Refactored Permissions into a single static file so that the strings aren't scattered all over the place.

TODO:
Refactor playwright tests in https://github.com/NHSDigital/nbs-appointments-management-service/pull/472 and merge in before this PR and use the new pattern.


src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.tsx
```diff
@@ -0,0 +1,129 @@
...
+        <TextInput
+          id="odsCode"
+          label="ODS code"
+          {...register('odsCode', {
+            required: true,
+          })}
+        ></TextInput>
+      </FormGroup>
+      <FormGroup
+        error={errors.icb ? 'You have not selected an ICB' : undefined}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.tsx:92]: (see other comment re:validation for these fields too)

src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.tsx
```diff
@@ -0,0 +1,129 @@
...
+      >
+        <Select
+          id="icb"
+          label="ICB"
+          options={mapWellKnownCodes('icb')}
+          {...register('icb', {
+            required: true,
+            validate: e => validateWellKnownCode(e, 'icb'),
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.tsx:100]: I'm all for the principle of failing fast, but this doesn't feel like it should be necessary. We've just set `options` to a map of this object so should know that their values are correct. I presume the backend has its own validation on this too if an erroneous value did get through

src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx
```diff
@@ -60,43 +61,70 @@ const EditDetailsForm = ({
...
+          })}
         ></TextInput>
         <TextInput
           id="longitude"
           label="Longitude"
-          {...register('longitude')}
+          {...register('longitude', {
+            required: true,
+            pattern: DECIMAL_REGEX,
+          })}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx:109]: So I got pulled up on stuff like this when we were working on the Create Availability screens. It was insisted upon at the time that users shouldn't be allowed to even type incorrect figures, hence why I had to switch everything over to controlled inputs. 

I kind of don't want to re-open that can of worms, but this is technically inconsistent. From reading the ticket it looks like they've just forgotten to scope out validation, but it'll get raised as a bug by the testing team at some point if we don't address it

src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx
```diff
@@ -60,43 +61,70 @@ const EditDetailsForm = ({
...
           id="phoneNumber"
           type="tel"
           label="Phone number"
-          title="Please enter numbers and spaces only."
-          pattern="[0-9 ]*"
-          {...register('phoneNumber')}
+          {...register('phoneNumber', {
+            required: true,
+            pattern: PHONE_NUMBER_REGEX,
+          })}
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx:127]: Likewise here, rather than pattern validation (which personally I'd prefer but nevermind), we'll have bugs raised against this if the users are allowed to type non-numeric characters

src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx
```diff
@@ -60,43 +61,70 @@ const EditDetailsForm = ({
 
   return (
     <form onSubmit={handleSubmit(submitForm)}>
-      <FormGroup error={errors.name?.message}>
+      <FormGroup
+        error={errors.name ? 'You have not entered a name' : undefined}
+      >
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx:66]: What's the reason for this change? Why not just set the error message during the validation?

src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+      name: 'Region One',
+    }) as HTMLOptionElement;
+    const option4 = screen.getByRole('option', {
+      name: 'Region Three',
+    }) as HTMLOptionElement;
+
+    expect(option1.selected).toBe(true);
+    expect(option2.selected).toBe(false);
+    expect(option3.selected).toBe(true);
+    expect(option4.selected).toBe(false);
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.test.tsx:77]: You could replace these with 
`expect(screen.getByRole('combobox', { name: 'ICB' })).toHaveDisplayValue('Integrated Care Board One');`

Would be clearer too as option1, option2, option 3 and option 4 sound like they all belong to the same control

src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx
Comment (jsed-nhs) [reply src/client/src/app/site/[site]/details/edit-details/edit-details-form.tsx:109]: Good point, I can ask about it, but might just stick to controlled inputs if you think thats the established pattern.

src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.tsx
Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.tsx:100]: If we don't trust what's in the db but we DO know how to validate it, we're then in a world where we should be writing validators for retrieval and validating the response from each GetFooFunction.cs. 

Obviously this is a huge red flag if we do this, but it's better than passing the problem along to the frontend

Comment (jsed-nhs) [reply src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.tsx:100]: To be fair it sounds like there is a ticket for validating the MYA Bulk Data now, so I'm not really sure where this leaves my work in terms of validation. Might need to chat to some BA's about it.

src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.test.tsx
Comment (jsed-nhs) [reply src/client/src/app/site/[site]/details/edit-reference-details/edit-reference-details-form.test.tsx:77]: I will add this display check in, however will just rename the existing Options one code above, as want to leave the HtmlOption selected true/false assertions in (protects against possible multi-select control)

sambiramairelogic (approved):

sambiramairelogic (approved):

PR #487, opened 2025-01-31
APPT-606 OKTA Invitation For Non-NHS Email

MikeHDigital:
This PR is wiring up the Okta user directory to provision users in Okta when they aren't using NHS Mail and don't currently exist in the Okta organisation.

Validating emails with a regex is a minefield, it was fine to check when it was just nhs emails and the new regex is likely good enough to match the vast majority of cases.


jsed-nhs:
Need to rename the flags in the environment feature flags.
Have a quick check if can put the PEM key somewhere more appropriate, but not blocking.

jsed-nhs:
Also, are all the environment variables setup against each environment? Dev/Stag/Prod? where are they stored, azure library/portal?

src/api/Nhs.Appointments.Api/Nhs.Appointments.Api.csproj
```diff
@@ -6,6 +6,7 @@
     <PlatformTarget>x64</PlatformTarget>
     <Nullable>disable</Nullable>
     <LangVersion>latestmajor</LangVersion>
+    <UserSecretsId>6c3b185b-e458-4b9d-9353-ba22192c1d1c</UserSecretsId>
```
jsed-nhs, line 9: Need to delete!

infrastructure/resources/high_load_functions.tf
```diff
@@ -174,6 +178,10 @@ resource "azurerm_windows_function_app_slot" "nbs_mya_high_load_func_app_preview
     Auth__Providers__1__ClientCodeExchangeUri                              = "${local.client_code_exchange_uri}?provider=okta"
     Auth__Providers__1__ReturnUri                                          = "${local.auth_provider_return_uri}?provider=okta"
     Auth__Providers__1__RequiresStateForAuthorize                          = true
```
jsed-nhs, line 180: Don't need these for high load functions

PR #494, opened 2025-02-04
Adding mechanism to validate tokens based on a configurable public key

MikeHDigital:
- Added a simple configurable mechanism to utilise a local public key (not from a jwks) to validate a token against
- Added a configuration flag to completely disable the addition of the local configurable key to ensure there is no ambiguity about the state of the authentication methods available
- Added an example console application to generate an rsa key pair and use the private side to sign a token that will be accepted by the token validator in the api.

For this to be usable the console app needs some work to take a list of the test users and output a file containing the user and corresponding signed token along with the public key which can be configured for an environment.

Currently these tokens are generated with a validity window of 10 years to demonstrate that the validity window can be whatever we need to facilitate continued load testing.


vicr1:
> * Added a simple configurable mechanism to utilise a local public key (not from a jwks) to validate a token against

I have a reservation about how this might affect performance testing. This represents a significant difference to production code, so it is a code change to support testing. The JWKS could be retrieved from a remote URI in the same way as it is for a real IDP and this would be more representative of the production mechanism and would give better indication of performance. Given that this code is involved in almost every request it might have a big impact. For example if a problem with the caching of JWKS was introduced it would be highlighted by a performance test, but not if the regular JWKS code is not circumvented

JwtGenerator/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+};
+
+var encodedPublicKey = Convert.ToBase64String(rsa.ExportRSAPublicKey());
+
+Console.WriteLine("PUBLIC KEY FOR CONFIG:");
+Console.WriteLine(encodedPublicKey);
+
+Console.WriteLine();
+
+// loop around different users
```
jsed-nhs, line 23: Unless I'm being blind I can't see a loop here!

JwtGenerator/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+var claims = new List<Claim>
+{
+    new(ClaimTypes.Name, "user1@user.com"),
```
jsed-nhs, line 26: Do we possibly need to add more claims? Email, Username etc? have you tested this locally and it works with just the Name claim?

jsed-nhs, line 26: Should we allow the running of this app to provide their own username/email in console args, so that you get a token for that specific user? Assume this one is just a single one for the draft PR?

jsed-nhs: Sorry just seen your PR description :D

JwtGenerator/Program.cs
jsed-nhs: Sorry it's in the PR description

vicr1 (changes requested):
As commented, I don't think we should be introducing code changes to support performance testing, especially as this would significantly change the behaviour of the system for performance test. It would be a simple matter to provide the JWKS via an endpoint (using Azure APIM) and hence no test only code would be required

jsed-nhs:
I think it largely depends on how deep you want this testing to go. I agree in trying to get the performance testing to be as close to the Production mechanism as possible - but there is a question of how far to take it and the dev effort/complexity involved. I was under the impression that most of the performance testing that this is satisfying is to do with load testing thousands of users and bookings over thousands of sites concurrently. 

Once this token is retrieved, it should be cached (same in Prod), so the mechanism to do the auth hand-shake (whilst not Production exact) shouldn't be being called on every API request, so I don't think the JWKS retrieval being skipped should have as large of an impact as you are implying? Happy to be wrong though, but I think that's why me and Mike agreed that this approach would be good enough for now.

If a Production issue arose in the scenario you are describing, I suppose so long as we're aware how our Staging Load Testing was done, we could narrow any live issues down through process of elimination - so long as we document that JWKS is skipped in Staging load testing.


> > * Added a simple configurable mechanism to utilise a local public key (not from a jwks) to validate a token against
> 
> I have a reservation about how this might affect performance testing. This represents a significant difference to production code, so it is a code change to support testing. The JWKS could be retrieved from a remote URI in the same way as it is for a real IDP and this would be more representative of the production mechanism and would give better indication of performance. Given that this code is involved in almost every request it might have a big impact. For example if a problem with the caching of JWKS was introduced it would be highlighted by a performance test, but not if the regular JWKS code is not circumvented

MikeHDigital:
So the code introduced isn't on the critical path for perf, namely that pulling a jwks is only done once per hour anyway, and this still ensures that the validation mechanism being executed for the test remains consistent. 

We can externally host a jwks but that will still require an additional auth provider configuration which is fundamentally equivalent to this piece of additional code, it will be adding a key and audience to the lists in this PR.

I'm happy to go with an external JWKS but offered this as it requires no change outside of the project and I don't presently have the ability to make or propose change to APIM.

vicr1:
> Once this token is retrieved, it should be cached (same in Prod), so the mechanism to do the auth hand-shake (whilst not Production exact) shouldn't be being called on every API request, so I don't think the JWKS retrieval being skipped should have as large of an impact as you are implying? Happy to be wrong though, but I think that's why me and Mike agreed that this approach would be good enough for now.

Should is the important word here, if there are any changes introduced that affect this then performance is one of the places where you might see it so removing the production code for performance test seems like the wrong approach. Especially as you can simply provide the JWKS from an endpoint and then no configuration change is needed other than to change the JWKS URI configuration item which is already an existing configuration item. This would then ensure that the production code paths are tested. Azure APIM can easily provide a static response for the JWKS

JwtGenerator/Program.cs
MikeHDigital: This is all that's used and works currently

MikeHDigital:
Additionally I dislike the idea of it being "just" another jwks configuration as I feel that poses additional risk in terms of potential production misconfiguration. This is a clear configuration which I feel presents less risk in terms of it accidentally being copied over to production by some future maintenance.

MikeHDigital:
I do appreciate the potential for a delay in jwks retrieval having a potential performance impact when it comes to processing this authentication however that process is still being executed as part of the token validation so any performance impact around the providers would still be expressed during the test runs, and this mechanism removes any potential for the additional test code or APIM being a source of those delays.

PR #523, opened 2025-02-16
APPT-627 - support querying availability by consecutive slots

pata9:
This change enables NBS to query availability by consecutive slots. This change allows NBS to pass 'consecutive' to the query availability endpoint.

The changes doesn't overly effect the endpoint point past slots. Anything above slots like daily will change the meaning of the count to say the amount of consecutive availability there is. When query by slot it will also return the consecutive slots but the logic is always based on consecutive looking forward. A slot that returns 0 when query by slots could be bookable because a slot before it has the calculation for that slot, but the 0 slot has no consecutive slots when looking forward.


pata9:
Need to find a good way of integration testing this

src/api/Nhs.Appointments.Core/SessionInstanceExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            return slots;
+        }
+
+        var consecutiveSlots = new List<SessionInstance>();
+
+        foreach (var slot in slots)
+        {
+            var timePeriod = new TimePeriod(slot.From, slot.Duration * consecutive);
+            var relatedCapacity = slots.Where(rs => rs.From >= timePeriod.From && rs.Until <= timePeriod.Until && slot.Services.Equals(rs.Services));
```
sambiramairelogic, line 20: How does this ensure appointments are consecutive? 
If time period was 10:00 -> 10:40 (e.g. Duration = 10, consecutive =4) and there was a 10:00 slot and a 10:30 slot, wouldn't they both be returned as "consecutive" slots with a gap between 10:10 and 10:30?

src/api/Nhs.Appointments.Api/Availability/QueryAvailabilityRequestValidator.cs
```diff
@@ -21,6 +21,10 @@ public QueryAvailabilityRequestValidator()
             .NotEmpty().WithMessage("Provide a valid string");
         RuleFor(x => x.QueryType)
             .Must(queryType => validQueryTypeValues.Contains(queryType)).WithMessage($"Value must be one of: {String.Join(", ", validQueryTypeValues)}");
+        RuleFor(x => x.Consecutive)
+            .NotEmpty()
+            .GreaterThanOrEqualTo(1)
+            .LessThanOrEqualTo(5).WithMessage("Consecutive must between 1 and 5");
```
sambiramairelogic, line 27: You could wrap these into one call using `.InclusiveBetween(1, 5)`

tests/Nhs.Appointments.Api.UnitTests/Functions/QueryAvailabilityFunctionTests.cs
jsed-nhs, line 1: Would surely want a lot of unit tests covering all manner of edge cases and availability setups to confirm the correct results were returned.

src/api/Nhs.Appointments.Core/SessionInstanceExtensions.cs
jsed-nhs: Agreed, would want these setups covering in tests also.

jsed-nhs:
Ideally would want a detailed description in the PR for what this is aiming to achieve, so that it can be used when referring back to if any changes required. (similar to test coverage)

jsed-nhs (dismissed):

src/api/Nhs.Appointments.Core/SessionInstanceExtensions.cs
pata9: I'll add a check to ensure that the duration of the availability is the same, I think this makes more sense than checking the services are the same

adam-bell-nhs:
Please can we make sure the new Consecutive param is not a breaking change (so nullable / defaulted to 1 internally) so NBS can continue to work until it's implemented on that side.

pata9:
> Please can we make sure the new Consecutive param is not a breaking change (so nullable / defaulted to 1 internally) so NBS can continue to work until it's implemented on that side.

@adam-bell-nhs this has been done in the PR, the api will support no Consecutive param and default to the value of 1. Going to add some tests around this, should be ready for you to test NBS side in the next couple of days :)

src/api/Nhs.Appointments.Api/Availability/QueryAvailabilityRequestValidator.cs
```diff
@@ -21,6 +21,9 @@ public QueryAvailabilityRequestValidator()
             .NotEmpty().WithMessage("Provide a valid string");
         RuleFor(x => x.QueryType)
             .Must(queryType => validQueryTypeValues.Contains(queryType)).WithMessage($"Value must be one of: {String.Join(", ", validQueryTypeValues)}");
+        RuleFor(x => x.Consecutive)
+            .InclusiveBetween(1, 5)
+            .WithMessage("Consecutive must between 1 and 5");
```
sambiramairelogic, line 26: Tiny typo here, need a "be" :)

sambiramairelogic: ight wanna say "between 1 and 5 inclusive" too, to be precise

src/api/Nhs.Appointments.Core/SessionInstanceExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        foreach (var slot in slots)
+        {
+            var timePeriod = new TimePeriod(slot.From, slot.Duration * consecutive);
+            var consecutivePeriods = generateConsecutivePeriods(slot, consecutive);
+            var consecutiveCapacity = slots.Where(rs => consecutivePeriods.Any(cp => cp.From == rs.From && cp.Until == rs.Until));
+
+            if (consecutiveCapacity.Count() <= 1)
+            {
+                continue;
+            }
```
sambiramairelogic, line 30: This is still only checking that there is at least 1 slot within the consecutive periods. It needs to check that every slot within the consecutive period is met by a matching availability slot. Sticking with this approach you'd need to change `if (consecutiveCapacity.Count() <= 1)` to `if (consecutiveCapacity.Count() != consecutivePeriods.Count())`, but I think there might be a less confusing way to write this

pata9: Thanks! I'll wrap that check in a method to solve the readability

src/api/Nhs.Appointments.Core/SessionInstanceExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+        if (consecutive <= 1)
+        {
+            return slots;
+        }
+
+        var consecutiveSlots = new List<SessionInstance>();
+
+        foreach (var slot in slots)
+        {
+            var consecutivePeriods = GenerateConsecutivePeriods(slot, consecutive);
```
pata9, line 19: this needs a rethink

src/api/Nhs.Appointments.Core/GroupSessionByConsecutive.cs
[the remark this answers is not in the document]
pata9: If we were to add this to the group it's adding more complexity to the grouping and tying the two concepts together. Not to mention having to check this feature flag in 3 places as there are at least 3 implementations of that grouper.

src/api/Nhs.Appointments.Api/Json/JsonToObjectSchemaValidation.cs
```diff
@@ -18,7 +18,7 @@ public static List<ErrorMessageResponseItem> ValidateConversion<TRequest>(string
             {
                 document = JsonDocument.Parse(json);
             }
-            catch(System.Text.Json.JsonException)
+            catch(System.Text.Json.JsonException ex)
```
sambiramairelogic, line 21: Is this change necessary?

pata9: No it's not, thanks for spotting, I'll double check my other debugging and make sure that hasn't creeped in as well :)

src/api/Nhs.Appointments.Core/GroupSessionByConsecutive.cs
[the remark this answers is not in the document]
sambiramairelogic: I don't agree - having two interfaces which look like they do the same thing is more complexity still. We should encapsulate grouping into just one. 

And you don't need to copy/paste the same code to use it in each grouper, simply inherit the method. Then the consecutive logic is still wrapped in its own class and testable on its own, but there isn't the extra overhead in the function itself

sambiramairelogic (approved):

PR #524, opened 2025-02-17
APPT-628 - confirm joint bookings concept added to confirm

pata9:
**ConfirmProvisionalBookingFunction** has been modified to accept multiple booking references:

- The same contact details are used for all bookings.
- The first booking reference (leadBooker) is saved against child bookings.
- This change does not affect the confirmation process for a single provisional booking.


jsed-nhs (changes requested):
Will need several Integration tests to lock down and document what this change is trying to accomplish

src/api/Nhs.Appointments.Persistance/Models/BookingIndexDocument.cs
```diff
@@ -23,4 +23,7 @@ public class BookingIndexDocument : IndexDataCosmosDocument
 
     [JsonProperty("status")]
     public AppointmentStatus Status { get; set; }
+
+    [JsonProperty("leadBooker")]
+    public string? LeadBooker { get; set; }
```
jsed-nhs, line 28: String nullable by default, can remove

jsed-nhs: In fact, does this LeadBooker even need to be in the index document? is it ever going to be queried? I can't see this property being set, or tested.

tests/Nhs.Appointments.Api.Integration/Scenarios/Booking/ConfirmBooking.feature
```diff
@@ -69,3 +69,97 @@
...
+    Then the call should be successful
+    And the bookings are no longer marked as provisional
+
+   Scenario: JB:Confirmation can record contact details
+     Given the site is configured for MYA
+     And the following sessions
+       | Date     | From  | Until | Services | Slot Length | Capacity |
+       | Tomorrow | 09:00 | 10:00 | COVID    | 5           | 1        |
+     And the following provisional bookings have been made
+       | Date     | Time  | Duration | Service |
```
jsed-nhs, line 98: These bookings will have the same booking reference in setup, is this intentional?

jsed-nhs (dismissed):
Leaving for now as not merging until R2. Will pick up again at a later point.

sambiramairelogic:
I'd like to bottom out this new concept of a "lead booker" before we commit it to the repo. That 628 ticket only says that we need an endpoint to accept plural provisional bookings, which could just be a flat array. It doesn't justify the need to create this relationship or define what it means. 

I've repeatedly been asking product/design what this means for rescheduling / orphaning / cancelling etc. Until we have business rules for this case I'm reticent to implement it tbh

pata9:
> I'd like to bottom out this new concept of a "lead booker" before we commit it to the repo. That 628 ticket only says that we need an endpoint to accept plural provisional bookings, which could just be a flat array. It doesn't justify the need to create this relationship or define what it means.
> 
> I've repeatedly been asking product/design what this means for rescheduling / orphaning / cancelling etc. Until we have business rules for this case I'm reticent to implement it tbh

@sambiramairelogic Lead booker is not a requirement, but a technical addition to enable us to retrospectively add management features like the ones you outline in future. It's been added for future proof purposes. Personally I didn't see too much value in adding this in but @MikeHDigital was keen that we kept ourselves flexible to future requirements

Dave-O-NHS (approved):

PR #525, opened 2025-02-17
APPT-343 - Expired booking cleanup bugfix

jsed-nhs:
Currently the cleanup process runs every 5 mins. This process deletes all provisional bookings that have been expired as of 5 mins ago. This creates an issue where, depending on when the initial booking was made, the 'expired' status of a provisional booking (5 mins old) only exists in the DB between 0-5 mins, before being hard-deleted. 

I have upped the cleanup expired time cleanup, so that now 'Expired' bookings (5 mins old) will exist in the DB for 10-15 more mins before being removed.

Tech Debt Future - We really should have any variables like this controlled via IOptionsMonitor and injected in via config, so we wouldn't have to make further code changes in future.


pata9 (changes requested):
Can we shove some config in around this?

pata9 (approved):

PR #536, opened 2025-02-19
APPT-640 Comment out OKTA login button

jsed-nhs:
Ideally would toggle this on/off using feature toggle management, but due to time constraints simpler to just hide it for now.


MikeHDigital (approved):
LGTM

PR #545, opened 2025-02-25
Appt 493 mya view availability auto tests

sarithalakkyreddy:
Added new auto test for view availability


src/client/testing/utils/date-utility.ts
```diff
@@ -17,3 +19,32 @@ export const getDateInFuture = (
...
+export const geRequiredtDateInFormat = (
+  numberOfDaysFromToday: number,
+  requiredformat: string,
+) => {
+  const date = new Date();
+  let requiredDate = '';
+  const futureDate = date.setDate(date.getDate() + numberOfDaysFromToday);
+  requiredDate = dayjs(futureDate).format(requiredformat);
+  return requiredDate;
+};
```
sambiramairelogic, line 32: This method could be a lot shorter and accept some default values, e.g. 
image

src/client/testing/utils/date-utility.ts
```diff
@@ -17,3 +19,32 @@ export const getDateInFuture = (
...
+
+export const getWeekRangeForRequiredDate = (requiredDate: string) => {
+  const date = dayjs(requiredDate).format('YYYY-MM-DD');
+  const startOfWeek = dayjs(date).startOf('isoWeek');
+  const endOfWeek = dayjs(date).endOf('isoWeek');
+  const formattedStartOfWeek = startOfWeek.format('D MMMM');
+  const formattedendOfWeek = endOfWeek.format('D MMMM');
+  return `${formattedStartOfWeek} to ${formattedendOfWeek}`;
+};
```
sambiramairelogic, line 50: These methods are identical apart from their first line, so we could refactor them to avoid this duplication, e.g. 
image

data/CosmosDbSeeder/items/local/booking_data/autotest_booking_01.json
```diff
@@ -0,0 +1,27 @@
+{
+  "reference": "autotest-booking-reference-01",
+  "from": "2025-08-05T09:30:00",
+  "duration": 10,
+  "service": "RSV:Adult",
+  "created": "2025-09-03T12:58:28.701582+00:00",
```
sambiramairelogic, line 6: It probably doesn't matter, but just to be extra safe can we make the Created date earlier than the From date, as a user wouldn't have been able to book an appointment in the past :)

src/client/testing/tests/view-availability/add-session.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+import MonthViewAvailabilityPage from '../../page-objects/view-availability-appointment-pages/month-view-availability-page';
+import { daysFromToday, weekHeaderText } from '../../utils/date-utility';
+import WeekViewAvailabilityPage from '../../page-objects/view-availability-appointment-pages/week-view-availability-page';
+import AddSessionPage from '../../page-objects/view-availability-appointment-pages/add-session-page';
+import AddServicesPage from '../../page-objects/view-availability-appointment-pages/add-services-page';
+import CheckSessionDetailsPage from '../../page-objects/view-availability-appointment-pages/check-session-details-page';
+import EditAvailabilityConfirmedPage from '../../page-objects/view-availability-appointment-pages/edit-availability-confirmed';
+import ChangeAvailabilityPage from '../../page-objects/view-availability-appointment-pages/change-availability-page';
+import CancelSessionDetailsPage from '../../page-objects/view-availability-appointment-pages/cancel-session-details-page';
+import DailyAppointmentDetailsPage from '../../page-objects/view-availability-appointment-pages/daily-appointment-details-page';
```
sambiramairelogic, line 15: I know this is dying for a refactor anyway (I'd like to introduce the pattern where clicking a button which triggers a page transition also returns the page object you've moved to, then we can do away with all these imports and don't have to new up 50 page objects on each test). 

Totally optional but I think we could add the page-objects folder to tsconfig so we can flatten these into a single import? Just for neatness' sake really :)

jsed-nhs: Am doing now :)

data/CosmosDbSeeder/items/local/index_data/autotest_booking_01_index.json
```diff
@@ -0,0 +1,10 @@
+{
+  "reference": "autotest-booking-reference-01",
```
sambiramairelogic, line 2: This reference (and id) want updating to match the now renamed booking 

Also I think we should add index documents for bookings 16 through 19 also

src/client/testing/page-objects/view-availability-appointment-pages/daily-appointment-details-page.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  async cancelAppointment(appointmentDetail: string) {
+    await this.page
+      .getByRole('row')
+      .filter({ hasText: `${appointmentDetail}` })
+      .getByRole('link', { name: 'Cancel' })
+      .click();
+  }
+
+  async confirmAppointmentCancelation(option: 'Yes' | 'No') {
```
sambiramairelogic, line 34: cancellation*

src/client/testing/page-objects/view-availability-appointment-pages/month-view-availability-page.ts
```diff
@@ -1,17 +1,122 @@
...
+  services: serviceOverview[];
+  totalAppointments: number;
+  booked: number;
+  unbooked: number;
+};
+
+export type serviceOverview = {
+  serviceName: string;
+  bookedAppointments: number;
+};
```
sambiramairelogic, line 15: Minor thing but elsewhere we're using PascalCase for types

Also I may have missed it, but are these used anywhere else? Might not need to be exported?

src/client/testing/page-objects/view-availability-appointment-pages/month-view-availability-page.ts
```diff
@@ -1,17 +1,122 @@
...
+    await this.page
+      .getByRole('listitem')
+      .filter({ has: this.page.getByText(`${requiredWeek}`) })
+      .getByRole('link', { name: 'View week' })
+      .click();
+  }
+
+  async navigateToRequiredMonth(month: string) {
+    await this.page.goto(
+      `/manage-your-appointments/site/6877d86e-c2df-4def-8508-e1eccf0ea6be/view-availability?date=${month}`,
```
sambiramairelogic, line 119: Could we parameterise this id? 
Again, in the future I'm envisaging page objects having a Site property that gets populated on transition but at least for now I think this should be supplied by the test itself

src/client/testing/page-objects/view-availability-appointment-pages/month-view-availability-page.ts
```diff
@@ -1,17 +1,122 @@
...
 
-  async verifyViewMonthDisplayed() {
+  async verifyViewNextMonthButtonDisplayed() {
     await expect(this.nextButton).toBeEnabled();
   }
+
+  async verifyAllWeekCardInformationDisplayedCorrectly(
+    expectedWeekOverviews: weekOverview[],
+  ) {
+    for (let i = 0; i < expectedWeekOverviews.length; i++) {
```
sambiramairelogic, line 38: I think this could be a foreach?

src/client/testing/page-objects/view-availability-appointment-pages/month-view-availability-page.ts
```diff
@@ -1,17 +1,122 @@
...
+      const weekOverview = expectedWeekOverviews[i];
+
+      const cardDiv = this.page
+        .getByRole('heading', {
+          name: weekOverview.header,
+        })
+        .locator('..');
```
sambiramairelogic, line 45: I'm not totally against stuff like this, but I think something like 
`this.page.getByRole('listitem').filter({ hasText: weekOverview.header });` would be neater? or `has: page.GetByRole('heading', weekOverview.header)` maybe

src/client/testing/page-objects/view-availability-appointment-pages/week-view-availability-page.ts
```diff
@@ -0,0 +1,250 @@
...
+
+export type dayOverview = {
+  header: string;
+  services: serviceOverview[];
+  totalAppointments: number;
+  booked: number;
+  unbooked: number;
+};
+
+export type serviceOverview = {
```
sambiramairelogic, line 12: Again, PascalCase and do they need to be exported?

src/client/testing/page-objects/view-availability-appointment-pages/week-view-availability-page.ts
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  async verifyViewNextWeekButtonDisplayed() {
+    await expect(this.nextButton).toBeEnabled();
+  }
+
+  async verifyAllDayCardInformationDisplayedCorrectly(
+    expectedDayOverviews: dayOverview[],
+  ) {
+    for (let i = 0; i < expectedDayOverviews.length; i++) {
```
sambiramairelogic, line 53: Same here, think this could be a foreach

src/client/testing/page-objects/view-availability-appointment-pages/week-view-availability-page.ts
```diff
@@ -0,0 +1,250 @@
...
+      const dayOverview = expectedDayOverviews[i];
+
+      const cardDiv = this.page
+        .getByRole('heading', {
+          name: dayOverview.header,
+        })
+        .locator('..');
```
sambiramairelogic, line 60: Likewise think we could use filter/has here to traverse the tree in just one direction rather than bouncing up

src/client/testing/page-objects/view-availability-appointment-pages/week-view-availability-page.ts
```diff
@@ -0,0 +1,250 @@
...
+          cardDiv.getByText(`Unbooked: ${dayOverview.unbooked}`, {
+            exact: true,
+          }),
+        ).toBeVisible();
+      } else {
+        //no services = no availability
+        await expect(cardDiv.getByText('No availability')).toBeVisible();
+
+        //no table
+        await expect(cardDiv.locator('table')).toHaveCount(0);
```
sambiramairelogic, line 139: Could this be getByRole rather than a locator?

src/client/testing/page-objects/view-availability-appointment-pages/week-view-availability-page.ts
```diff
@@ -0,0 +1,250 @@
...
+    await expect(
+      this.page
+        .getByRole('listitem')
+        .filter({ has: this.page.getByText(`${requiredDate}`) })
+        .getByText(`${service}`)
+        .first(),
+    ).toBeVisible();
+  }
+
+  async openDailyAppoitmentPage(appointmentDate: string) {
```
sambiramairelogic, line 243: typo in appointment

src/client/testing/tests/view-availability/add-session.spec.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  await weekViewAvailabilityPage.verifySessionAdded();
+  await weekViewAvailabilityPage.openDailyAppoitmentPage(requiredDate);
+  await dailyAppointmentDetailsPage.verifyDailyAppointmentDetailsPageDisplayed();
+});
+
+test('Verify appointment not cancelled when not confirmed', async ({
+  page,
+}) => {
+  const requiredDate = '2025-08-05';
```
sambiramairelogic, line 256: Any reason this one is hardcoded when others in this file use `daysFromToday` ?

src/client/testing/page-objects/view-availability-appointment-pages/week-view-availability-page.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (count == 1) {
+      await addAvailabilityButton.click();
+    } else {
+      await this.page
+        .getByRole('listitem')
+        .filter({ has: this.page.getByText(`${requiredDate}`) })
+        .getByRole('link', { name: 'Add Session' })
+        .click();
+    }
+  }
```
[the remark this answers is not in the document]
jsed-nhs: It doesn't theres a race condition where because the tests all run in parralell, it is not consistent as to whether the day has an existing session or not based on other test runs. Solution is to either keep parralel and do this (able to handle either case) or make the tests serial (slowing down the test run) but making it more expected.

src/client/testing/page-objects/view-availability-appointment-pages/month-view-availability-page.ts
jsed-nhs: The cards aren't list items, they are just divs?

sambiramairelogic: Yes but they all live inside listitem elements and you can guarantee 1 list item per card, so it's the same difference

src/client/testing/page-objects/view-availability-appointment-pages/week-view-availability-page.ts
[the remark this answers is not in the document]
sambiramairelogic: Could each test look for a different day so there aren't any data clashes?

src/client/testing/tests/view-availability/add-session.spec.ts
jsed-nhs: I believe it depends on a specific hard coded seed booking (20)

sambiramairelogic (approved):

PR #546, opened 2025-02-26
APPT-678: Surface Build Number in frontend

sambiramairelogic:
Surfaces a Build Number in the footer, hidden both visually and from the accessibility tree. 

This build number is currently just the pipeline ID - we can come up with another scheme if we want


sambiramairelogic:
As expected, have proved Playwright is using the build number:
image

pata9 (approved):
We may want to bring in a more meaningful scheme later, but I think we can role with this for now

PR #547, opened 2025-02-26
APPT-655 - Added cache control headers

Dave-O-NHS:
A new middleware has been implemented to add multiple headers to HTTP responses. 
These headers are designed to prevent Akamai from caching API responses.

### DIFF
--- a/src/api/Nhs.Appointments.Api/Middleware/NoCacheMiddleware.cs
+++ b/src/api/Nhs.Appointments.Api/Middleware/NoCacheMiddleware.cs
@@ -1,0 +1,16 @@
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Middleware;
+using System.Threading.Tasks;
+
+namespace Nhs.Appointments.Api.Middleware;
+public class NoCacheMiddleware : IFunctionsWorkerMiddleware
+{
+    public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+    {
+        await next(context);
+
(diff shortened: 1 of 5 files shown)


sambiramairelogic (approved):

src/api/Nhs.Appointments.Api/Middleware/NoCacheMiddleware.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Nhs.Appointments.Api.Middleware;
+public class NoCacheMiddleware : IFunctionsWorkerMiddleware
+{
+    public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+    {
+        await next(context);
+
+        var response = context.GetHttpContext().Response;
+        response.Headers["Cache-Control"] = "no-store, no-cache, max-age=0";
+        response.Headers["Pragma"] = "no-cache";
```
pata9, line 14: Apparently this deprecated? I don't think we need to be backwards compatible with http 1.0? 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma

Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

pata9 (approved):

PR #563, opened 2025-03-05
Fixing notifications for new user's roles - previously weren't sending

ste-banks:


ste-banks:
Closing as I'm going to be making changes in another branch that should handle this.

PR #566, opened 2025-03-06
DRAFT: Multi-service availability calculations

sambiramairelogic:
To enable Multi-Service Availability Calculations, we need to introduce new methods to the availability and bookings services. 

This PR creates those new methods and locks them behind a feature toggle, toggled off in all cases (as this feature isn't even ready in local). 

This PR also introduces the concept of a test class attribute we can use to run entire suites with or without certain toggles.


src/api/Nhs.Appointments.Core/AvailabilityService.cs
```diff
@@ -96,4 +152,147 @@ private static IEnumerable<DateOnly> GetDatesBetween(DateOnly start, DateOnly en
...
+            .FirstOrDefault();
+
+    public async Task<AvailabilityState> GetAvailabilityState(string site, DateOnly day)
+    {
+        var availabilityState = new AvailabilityState();
+
+        var orderedLiveBookings = await GetOrderedLiveBookings(site, day);
+        var slots = await GetSlots(site, day);
+
+        using var leaseContent = siteLeaseManager.Acquire(site);
```
pata9, line 173: I don't think we need to lease in a get state? Might be miss understanding the lease

src/api/Nhs.Appointments.Core/AvailabilityService.cs
```diff
@@ -96,4 +152,147 @@ private static IEnumerable<DateOnly> GetDatesBetween(DateOnly start, DateOnly en
...
+        availabilityState.AvailableSlots = slots.Where(s => s.Capacity > 0).ToList();
+        availabilityState.Bookings = orderedLiveBookings;
+
+        return availabilityState;
+    }
+
+    public async Task<AvailabilityState> RecalculateAppointmentStatuses(string site, DateOnly day)
+    {
+        var availabilityState = await GetAvailabilityState(site, day);
+
```
pata9, line 215: I think this is where the lease should live?

pata9 (approved):

PR #568, opened 2025-03-07
APPT-699 - Returning the report if any row fails validation before doing extra checks

ste-banks:


pata9 (changes requested):
What's the rational?

pata9 (approved):

sambiramairelogic (approved):

PR #570, opened 2025-03-10
Appt 672 Feature Toggle Management - Azure and Local Settings

jsed-nhs:
Ability to apply Feature Toggles with FeatureFilters using new Azure App Configuration resource.
The Azure App config is registered with a default 10 minute cache to prevent racking up too many requests on the resource, this means that any changes made to the config will take up to 10 mins to apply.
The supported Feature Filters available include:

- Simple On/Off
- Random % rollout for all
- Enabled within a pre-defined time window.
- Enabled for specific users
- Enabled for specific sites

Register AzureAppConfigManagement in ServiceRegistration, or use local.feature.flags.json if in local run mode.
Add SDK Microsoft.FeatureManagement.AspNetCore
Add FeatureToggleHelper class that adds all the user and site context into the FeatureManager enabled check.
Collection of test flags defined and how they might be used.
Add unit tests that confirm FeatureToggleHelper applies the user and site to the context correctly.


src/api/Nhs.Appointments.Api/local.settings.json
```diff
@@ -12,6 +12,7 @@
     "FUNCTIONS_WORKER_RUNTIME": "dotnet-isolated",
     "COSMOS_ENDPOINT": "https://localhost:8081",
     "COSMOS_TOKEN": "C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==",
+    "AzureAppConfigConnection": "local",
```
jsed-nhs, line 15: This is the setting that will be swapped out per environment for that envs app config

src/api/Nhs.Appointments.Api/Features/Flags.cs
```diff
@@ -0,0 +1,21 @@
+namespace Nhs.Appointments.Api.Features;
+
+public static class Flags
+{
+    #region TestFlags
```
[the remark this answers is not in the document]
jsed-nhs: Ahh okay, was just trying to group them up, so when people add real ones in they aren't alongside them. Can remove if you like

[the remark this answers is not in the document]
sambiramairelogic: What do you mean real ones? Are these not used? If so shouldn't they be in loca.feature.flags.json or something? Not sure quite I'm understanding the intention here

src/api/Nhs.Appointments.Api/Features/FeatureToggleHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+public class FeatureToggleHelper(IFeatureManager featureManager) : IFeatureToggleHelper
+{
+    public async Task<bool> IsFeatureEnabledForFunction(string featureFlag, FunctionContext functionContext,
+        ClaimsPrincipal principal, IRequestInspector requestInspector)
+    {
+        var siteIds = (await requestInspector.GetSiteIds(await functionContext.GetHttpRequestDataAsync())).ToArray();
+        return await IsFeatureEnabled(featureFlag, principal.Claims.GetUserEmail(), siteIds);
+    }
+
+    public async Task<bool> IsFeatureEnabled(string featureFlag, string userId, string[] sites)
```
sambiramairelogic, line 23: So how does this work if a feature is enabled for some sites but not all?

src/api/Nhs.Appointments.Api/Functions/GetFeatureFlagFunction.cs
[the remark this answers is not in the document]
jsed-nhs: Primarily debugging for devs/testers to check a flag on a specific environment, but could be used by the front end for example to fetch the state of a flag. In this case, it could be refactored if not exactly right.

src/api/Nhs.Appointments.Api/Features/Flags.cs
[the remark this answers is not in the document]
jsed-nhs: They do exist in the local.flags.json. This class is just a static c# definition list of flags in the system, as when you check you tend to use the string. Defining them here is just a nice way of making sure the string can't accidentally be misspelled in multiple places. The strings listed here ideally would match all in the local settings or Azure config.

src/api/Nhs.Appointments.Api/Features/FeatureToggleHelper.cs
jsed-nhs: It works on a logical OR. I have added tests in to prove this for a flag that is enabled for sites and users.

src/api/Nhs.Appointments.Api/Functions/GetFeatureFlagFunction.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Attributes;
+using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Extensions;
+using Microsoft.Extensions.Logging;
+using Nhs.Appointments.Api.Features;
+using Nhs.Appointments.Api.Models;
+using Nhs.Appointments.Core;
+
+namespace Nhs.Appointments.Api.Functions;
+
+public class GetFeatureFlagFunction(
```
pata9, line 18: I think we want to have a single source of truth for feature flags, so the UI will request feature flags from the app configuration? Can remove this?

src/api/Nhs.Appointments.Api/Functions/GetFeatureFlagsBySiteFunction.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Attributes;
+using Microsoft.Extensions.Logging;
+using Nhs.Appointments.Api.Features;
+using Nhs.Appointments.Api.Models;
+using Nhs.Appointments.Core;
+using Nhs.Appointments.Core.Inspectors;
+
+namespace Nhs.Appointments.Api.Functions;
+
+public class GetFeatureFlagsBySiteFunction(
```
pata9, line 19: Same as before here, can remove?

src/api/Nhs.Appointments.Api/Functions/GetFeatureFlagsByUserFunction.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Attributes;
+using Microsoft.Extensions.Logging;
+using Nhs.Appointments.Api.Features;
+using Nhs.Appointments.Api.Models;
+using Nhs.Appointments.Core;
+using Nhs.Appointments.Core.Inspectors;
+
+namespace Nhs.Appointments.Api.Functions;
+
+public class GetFeatureFlagsByUserFunction(
```
pata9, line 19: And again here can remove?

src/api/Nhs.Appointments.Api/Functions/GetFeatureFlagFunction.cs
jsed-nhs: These 3 functions were being used by the postman endpoints to return flag states for debugging/test purposes. I will liaise with test and ask which they would like to keep, and what format they would like them in.

pata9 (approved):

pata9 (dismissed):

pata9 (dismissed):

PR #583, opened 2025-03-17
APPT-705 okta bulk upload

Dave-O-NHS:
Added validation for bulk uploaded okta users. FirstName and LastName is required.

### DIFF
--- a/src/api/Nhs.Appointments.Core/UserImportRowMap.cs
+++ b/src/api/Nhs.Appointments.Core/UserImportRowMap.cs
@@ -1,3 +1,4 @@
+using CsvHelper;
 using CsvHelper.Configuration;
 using static Nhs.Appointments.Core.UserDataImportHandler;
 
@@ -18,6 +19,12 @@ public class UserImportRowMap : ClassMap<UserImportRow>
         Map(m => m.UserId)
             .Name("User")
             .Validate(f => !string.IsNullOrWhiteSpace(f.Field));
+        Map(m => m.FirstName)
+            .Name("FirstName")
+            .Validate(f => ValidateName(f.Row, "FirstName"));
+        Map(m => m.LastName)
+            .Name("LastName")
+            .Validate(f => ValidateName(f.Row, "LastName"));
         Map(m => m.SiteId)
             .Name("Site")
             .TypeConverter<GuidStringTypeConverter>();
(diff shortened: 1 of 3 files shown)


src/api/Nhs.Appointments.Core/UserImportRowMap.cs
```diff
@@ -18,6 +19,12 @@ public UserImportRowMap()
         Map(m => m.UserId)
             .Name("User")
             .Validate(f => !string.IsNullOrWhiteSpace(f.Field));
+        Map(m => m.FirstName)
+            .Name("FirstName")
+            .Validate(f => ValidateName(f.Row, "FirstName"));
```
pata9, line 24: I think we can use `When(IsOktaUser)` here

pata9 (changes requested):

src/api/Nhs.Appointments.Core/UserImportRowMap.cs
Dave-O-NHS: We could, do you see any benefits of doing it? With current implementation, we only running validation for okta users.

Dave-O-NHS: I had a look, there is no .Where method available. 
image
I was trying to use .Convert and return null for non okta rows, but returning null disables validation method so this does not work. 
I think it would be the best to stick with current implementation.

pata9 (approved):

src/api/Nhs.Appointments.Core/UserImportRowMap.cs
```diff
@@ -35,4 +42,20 @@ public UserImportRowMap()
...
+    }
+
+    private bool ValidateName(IReaderRow row, string fieldName)
+    {
+        if (IsOktaUser(row))
+        {
+            var name = row.GetField(fieldName);
+            return !string.IsNullOrWhiteSpace(name);
+        }
+        return true; // No validation required for non-okta users
```
sambiramairelogic, line 59: Is there any reason we're not validating nhs.net users? Since we're just checking for null or whitespace, I don't see why we shouldn't apply this to all email addresses

Dave-O-NHS: Because we do not keep/require firstName or lastName for nhs.net users, we only need them for okta users.  
We already validate all email addressed on line 21.

sambiramairelogic: Ah, my apologies, I misunderstood. Okay looks good, though it does seem a shame that ClassMap library doesn't provide a better way of doing this

sambiramairelogic (approved):

PR #586, opened 2025-03-17
resolve case sensitive checks around user email

pata9:
Bugs resolved include - APPT-719, APPT-720, APPT-691


sambiramairelogic (changes requested):
I think we should add some regression tests around this behaviour or this change could easily get lost

pata9:
> I think we should add some regression tests around this behaviour or this change could easily get lost

Added a couple of tests around this

PR #588, opened 2025-03-18
terraform app configuration

pata9:


infrastructure/environments/dev/main.tf
```diff
@@ -17,7 +17,12 @@ terraform {
 }
 
 provider "azurerm" {
-  features {}
+  features {
+    app_configuration {
```
jsed-nhs, line 21: We removing these?

jsed-nhs (changes requested):

infrastructure/environments/dev/main.tf
pata9: Nah we need to disable them to stop the terraform making the check

pata9: they default to true, I'm setting them to false

jsed-nhs (dismissed):

jsed-nhs (approved):

PR #595, opened 2025-03-21
APPT-689 - Adding accessibility statement to the footer

ste-banks:


src/client/src/app/lib/components/nhs-footer.tsx
```diff
@@ -25,6 +25,10 @@ const NhsFooter = () => {
           text: 'Cookies Policy',
           href: '/cookies-policy',
         },
+        {
+          text: 'Accessibility Statement',
+          href: 'https://www.nhs.uk/our-policies/manage-your-appointment-accessibility-statement',
```
sambiramairelogic, line 30: This link 404s - I thought https://nhsd-jira.digital.nhs.uk/browse/APPT-688 needed to be done first, then we were going to add a new page into MYA that this would point to?

ste-banks: Yeah I mentioned that to Jack, he said it's expected it won't work yet and will when the page goes live. I'm not sure what the status around APPT-688 is but it was mentioned in stand up on Wednesday that this one was ready to be developed.

sambiramairelogic: Fine, but on their heads be it if that ticket doesn't get done in time and this gets released into production...

PR #600, opened 2025-03-21
Add a 1-minute refresh interval for feature flags. This does not poll…

jsed-nhs:
… the app config when idle.


sambiramairelogic:
What exactly do you mean by "when idle" in the context of a function app?

PR #603, opened 2025-03-24
APPT-672 Add registration of IConfigurationRefresher.

jsed-nhs:
The app has the ability to call the refresh when a feature toggle state is requested. Fire off without awaiting to avoid execution delay. Register local config refresher. Refactor all feature related registration into a single place.

### DIFF
--- a/src/api/Nhs.Appointments.Api/Features/FeatureToggleHelper.cs
+++ b/src/api/Nhs.Appointments.Api/Features/FeatureToggleHelper.cs
@@ -1,11 +1,12 @@
 using System;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
 using Microsoft.FeatureManagement;
 using Microsoft.FeatureManagement.FeatureFilters;
 
 namespace Nhs.Appointments.Api.Features;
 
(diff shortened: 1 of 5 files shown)


src/api/Nhs.Appointments.Api/Features/FeatureToggleHelper.cs
```diff
@@ -14,6 +15,10 @@ public async Task<bool> IsFeatureEnabled(string featureFlag)
             throw new ArgumentException("FeatureFlag cannot be null or empty.");
         }
         
+        // fire and forget to not block execution
+        // means a slight delay in applying the very latest configuration, but its worth it to not block every time this is invoked
+        _ = configRefresher.RefreshAsync();
```
pata9, line 20: This could be called multiple times per request? Could shove it in middleware to guarantee it's called once per transient scope?

pata9 (approved):

src/api/Nhs.Appointments.Api/Features/FeatureToggleHelper.cs
jsed-nhs: Could add it to middleware in the future if the pattern ends up requiring it.

PR #606, opened 2025-03-27
APPT-733 - Hiding the bulk import feature behind a feature flag

ste-banks:


src/api/Nhs.Appointments.Api/Functions/BulkImportFunction.cs
```diff
@@ -44,17 +46,23 @@ public override Task<IActionResult> RunAsync(
...
-    protected override Task<(IReadOnlyCollection<ErrorMessageResponseItem> errors, BulkImportRequest request)>
+    protected override async Task<(IReadOnlyCollection<ErrorMessageResponseItem> errors, BulkImportRequest request)>
         ReadRequestAsync(HttpRequest req)
     {
-        var files = req.Form?.Files;
         var errors = new List<ErrorMessageResponseItem>();
 
+        if (!await featureToggleHelper.IsFeatureEnabled(Flags.BulkImport))
+        {
+            errors.Add(new ErrorMessageResponseItem { Message = "Bulk import feature is not enabled." });
```
pata9, line 56: I think just return 404? What do you think?

ste-banks: Yeah I do think a 404 would be better. Initially have done it like this as the ReadRequestAsync method can't currently return a 404 so I'm not sure if we'd need a all round API feature flag solution we can use for that? Happy to have a discussion with the devs around it

ste-banks: Have updated it to return a 404 now instead.

sambiramairelogic (approved):

PR #609, opened 2025-03-28
Introduce Multi-Service Opportunity Cost Algorithm

sambiramairelogic:
Extends the new Multi-Service Availability Calculations to use Jonny's algorithm. 

We have 2 tests which were proving the old code fell foul of the Best Fit problem, these tests have been updated in this PR to prove the new code fixes that problem.


pata9 (dismissed):

sambiramairelogic:
This work is superseded by multiple more recent branches

PR #615, opened 2025-04-02
APPT-615 - Adding OKTA users to OKTA as well via the bulk import

ste-banks:
Also removing the commented out permission check on the bulk import


pata9:
> > Should we be checking the okta feature toggle here somewhere?
> 
> We do in the UserImportRowMap. It's checks if a user is an okta user and if they are & okta is disabled, it will be added to the report and returned without any users being added.

You alright to add a test for this error?

ste-banks:
> > > Should we be checking the okta feature toggle here somewhere?
> > 
> > 
> > We do in the UserImportRowMap. It's checks if a user is an okta user and if they are & okta is disabled, it will be added to the report and returned without any users being added.
> 
> You alright to add a test for this error?

There already is one, it's part of the UserDataImportHandlerTests - David added it as part of his ticket.

pata9 (approved):

PR #621, opened 2025-04-03
APPT-763 - order blocks for slots and update tests to assert the order is correct

pata9:


ste-banks (approved):

src/api/Nhs.Appointments.Api/Availability/SlotAvailabilityGrouper.cs
sambiramairelogic, line 1: I agree with this change, but I think we should also make it in the `HourlyAvailabilityGrouper`. Unless I'm mistaken that could return hours out of sequence as well

sambiramairelogic (changes requested):
I agree with this change, but I think we should also make it in the HourlyAvailabilityGrouper. Unless I'm mistaken that could return hours out of sequence as well

sambiramairelogic (dismissed):

PR #626, opened 2025-04-07
APPT-765 - Calculating the bookings from midnight of the day after the weekend

ste-banks:
I've not update tests for this as Jonny is currently writing a comprehensive test suite around timings in his branch and we agreed that this change will be covered as part of it. On top of that, Saritha has a PR for View Availability tests which cover this area as well.

### DIFF
--- a/src/client/src/app/lib/services/availabilityCalculatorService.ts
+++ b/src/client/src/app/lib/services/availabilityCalculatorService.ts
@@ -32,7 +32,7 @@ export const summariseWeek = async (
     ),
     fetchBookings({
       from: weekStart.format('YYYY-MM-DD HH:mm:ss'),
-      to: weekEnd.add(1, 'day').endOf('day').format('YYYY-MM-DD HH:mm:ss'),
+      to: weekEnd.endOf('day').format('YYYY-MM-DD HH:mm:ss'),
       site: siteId,
     }),
   ]);


src/client/src/app/lib/services/availabilityCalculatorService.ts
```diff
@@ -32,7 +32,7 @@ export const summariseWeek = async (
     ),
     fetchBookings({
       from: weekStart.format('YYYY-MM-DD HH:mm:ss'),
-      to: weekEnd.format('YYYY-MM-DD HH:mm:ss'),
+      to: weekEnd.add(1, 'day').format('YYYY-MM-DD HH:mm:ss'),
```
sambiramairelogic, line 35: Could we do this with `.endOf('day')` instead? Not sure if it would give you midnight or 23:59.99...

ste-banks: Will update it and give it a quick test, 23:59:59 should be grand

ste-banks: Updated

src/client/src/app/lib/services/availabilityCalculatorService.ts
```diff
@@ -32,7 +32,7 @@ export const summariseWeek = async (
     ),
     fetchBookings({
       from: weekStart.format('YYYY-MM-DD HH:mm:ss'),
-      to: weekEnd.format('YYYY-MM-DD HH:mm:ss'),
+      to: weekEnd.add(1, 'day').endOf('day').format('YYYY-MM-DD HH:mm:ss'),
```
sambiramairelogic, line 35: Do we still need to add a day? I thought the issue was that weekEnd is midnight at the start of Sunday, which would now give us 23:59 the next Monday

Sorry if I'm wrong and weekEnd is actually Saturday

sambiramairelogic (approved):

src/client/src/app/lib/services/availabilityCalculatorService.ts
ste-banks: You're right. I tested it with both then didn't remove the add day - have removed now

PR #631, opened 2025-04-08
make user role change only return its scope so that notifications dont consider roles that arent relevant as they are not part of the scope

pata9:
The notifications are being sent with the wrong information. This is due to Users having the same roles at different sites. This changes makes the store works in a single scope, returning only the original roles that are relevant for change (scope).

### DIFF
--- a/src/api/Nhs.Appointments.Persistance/UserStore.cs
+++ b/src/api/Nhs.Appointments.Persistance/UserStore.cs
@@ -35,7 +35,7 @@ public class UserStore(ITypedDocumentCosmosStore<UserDocument> cosmosStore, IMap
     /// <param name="roleAssignments"></param>
     /// <returns>The original role assignments for the scope prior to completion of the update operation</returns>
     /// <exception cref="Exception"></exception>
-    public async Task<RoleAssignment[]> UpdateUserRoleAssignments(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
+    public async Task UpdateUserRoleAssignments(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
     {
         var originalDocument = await GetOrDefaultAsync(userId);
         if (originalDocument == null)
@@ -46,7 +46,6 @@ public class UserStore(ITypedDocumentCosmosStore<UserDocument> cosmosStore, IMap
                 RoleAssignments = roleAssignments.ToArray()
             };
             await InsertAsync(user);
-            return [];
         }
         else
         {
@@ -57,8 +56,6 @@ public class UserStore(ITypedDocumentCosmosStore<UserDocument> cosmosStore, IMap
                 .Concat(roleAssignments);
             var userDocumentPatch = PatchOperation.Add("/roleAssignments", newRoleAssignments);
             await cosmosStore.PatchDocument(documentType, userId, userDocumentPatch);
-
-            return originalDocument.RoleAssignments.Where(x => x.Scope == scope).ToArray();
         }
     }
 
--- a/src/api/Nhs.Appointments.Persistance/UserStore.cs
+++ b/src/api/Nhs.Appointments.Persistance/UserStore.cs
@@ -1,4 +1,4 @@
-﻿using AutoMapper;
+using AutoMapper;
 using Microsoft.Azure.Cosmos;
 using Nhs.Appointments.Core;
 using Nhs.Appointments.Persistance.Models;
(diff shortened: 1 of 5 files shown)


tests/Nhs.Appointments.Persistance.UnitTests/UserStoreTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        private readonly Mock<IMapper> _mapper = new();
+        private readonly UserStore _sut;
+
+        public UserStoreTests()
+        {
+            _sut = new UserStore(
+                _cosmosStore.Object,
+                _mapper.Object
+                );
+        }
```
jsed500, line 19: Do we need to add tests for the SaveUserAsync method if missing and you've altered it?

src/api/Nhs.Appointments.Persistance/UserStore.cs
```diff
@@ -58,33 +58,13 @@ public async Task<RoleAssignment[]> UpdateUserRoleAssignments(string userId, str
...
 
-
-            return originalDocument.RoleAssignments;
+            return originalDocument.RoleAssignments.Where(x => x.Scope == scope).ToArray();
         }
     }
 
     public async Task SaveUserAsync(string userId, string scope, IEnumerable<RoleAssignment> roleAssignments)
     {
-        var originalDocument = await GetOrDefaultAsync(userId);
```
jsed500, line 68: Will the patch on its own work, does it not still need to try and create the user doc if not exists?

ste-banks: The code looks to be duplicated between the two methods so it _should_ be fine.

jsed500: Yes, if the patch method alone works, merge into one method and call it upsert or something

pata9: Ye it's duplicated code, that's why I changed it, looks like the only difference is that it doesn't return the roles? I decided to just make save call this method instead of doing a larger refactor

tests/Nhs.Appointments.Persistance.UnitTests/UserStoreTests.cs
pata9: They're the same tbh, won't really add much value, just adding more to be refactored. I just wanted to test my change, it doesn't effect the SaveUserAsync..

src/api/Nhs.Appointments.Persistance/UserStore.cs
jsed-nhs: Ahh okay didn't notice that difference. Happy to leave as two methods if you want.

tests/Nhs.Appointments.Persistance.UnitTests/UserStoreTests.cs
jsed-nhs: I think even though the implementation is almost identical, would be still nice to have tests for the Save method (even if they are mostly duplicated, as that acts as documentation if someone wants to change the Save method in future)

src/api/Nhs.Appointments.Persistance/UserStore.cs
```diff
@@ -33,7 +33,7 @@ public async Task<IEnumerable<RoleAssignment>> GetUserRoleAssignments(string use
     /// <param name="userId"></param>
     /// <param name="scope"></param>
     /// <param name="roleAssignments"></param>
-    /// <returns>The original role assignments prior to completion of the update operation</returns>
+    /// <returns>The original role assignments for the scope prior to completion of the update operation</returns>
```
sambiramairelogic, line 36: I know this isn't a change you've introduced, but this really jarred me when I first read it. I would expect this to either return nothing at all, or the user's roles as they NOW are, after the update has been applied. 
Looking at the code, it seems we're relying on this method giving us old info to perform the following logic: 
image

sambiramairelogic: Part of me wonders if it'd be better to keep this method uncomplicated, and filter by scope in the block above. That keeps a better separation of concerns. Equally though, I don't think we should be using this method to derive the "old" roles anyway so... 🤷

src/api/Nhs.Appointments.Persistance/UserStore.cs
pata9: I've pushed an a change it wasn't much to refactor

PR #633, opened 2025-04-10
Appt 662/dont show expired provisional bookings

ste-banks:


pata9 (approved):

jsed-nhs:
Expect this will be subject to more Timezone logic, agreed best to try and solve this after I've got my UTC changes into main, then readdress this

PR #639, opened 2025-04-11
prod okta config passed in terraform

pata9:


DACA16 (approved):
Approving, we might want to consider in future naming consistently, will discuss next week
okta_domain = var.OKTA_ISSUER

PR #638, opened 2025-04-11
APPT-749 Refactor all datetime dayJs issues to decouple from server/client timezones.

jsed-nhs:
**Background**

This pull request resolves a series of long-standing issues related to inconsistent datetime handling across the frontend of MYA. These inconsistencies primarily stemmed from discrepancies between the server timezone (UTC) and the user's browser timezone (typically Europe/London). Such differences resulted in unreliable behaviour, particularly around daylight saving time (DST) transitions, and were often mitigated through brittle workarounds that compounded the problem.

**Objective**
This work provides a long-term solution that decouples all frontend datetime logic from both server and browser timezones. The goal is to eliminate future issues of this nature, provided developers adopt the new patterns and utilities introduced here.

**Key Changes**

- Datetime Logic Abstraction
All direct use of dayjs has been removed in favour of the refactored timeService. This ensures consistent datetime creation, manipulation, and comparison throughout the application.

- Enforcement via Linting
A custom linting rule has been added to disallow direct imports of dayjs, encouraging exclusive use of timeService methods.

- Timezone Standardisation
All dayjs objects are now created and manipulated in the Europe/London timezone.

- Storage Format
Session datetimes are now stored as UK datetime strings rather than UTC dayjs objects. These strings are parsed into UK datetimes as needed, removing the use of UTC dayJs objects.

- UTC for Comparison Only
Datetime comparisons now convert both operands to UTC to work around known issues in dayjs. This is the only remaining use of UTC within the frontend.

- Custom Utilities
Several new helper methods have been added to timeService to address known edge cases:

addToUkDatetime: Replaces dayjs.add, accounting for DST boundary conditions.
isOnTheSameUkDay: A reliable alternative to dayjs.isSame for checking if two datetimes fall on the same UK day.
ukNow: Returns the current datetime in the UK timezone.
Constants for dateFormat and dateTimeFormat have been introduced for consistency.

- Refactoring
Several existing methods have been renamed or refactored to improve clarity and standardise usage patterns.

- Locale Updates
The en locale has been added, and all 'isoWeek' usages have been replaced with 'week' to prevent UTC-related complications.

**End-to-End Testing Enhancements**

- Extensive test coverage has been added to validate behaviour around DST boundaries:
- Seeded sessions and bookings for 25–27 October 2025 and 28–30 March 2026.
- Bookings deliberately placed at the edge of session times to ensure inclusion.

Frontend timezone robustness has been verified by running the application and its tests across a combination of timezones for the node server and client timezones:

Europe/London
[...]

PR #649, opened 2025-04-24
APPT-749 Fix dateTime issue when the API is ran in a different timezone

jsed-nhs:
Small missed edge case from me with the previous PR https://github.com/NHSDigital/nbs-appointments-management-service/pull/638 

When running the API in a different timezone also, I wanted to double check that the comprehensive combination of node timezone and the session tests still passed.

I noticed that the DailyAppointments page fails the tests in this case. This is because it passes the datetime info into the booking query to the API in UTC format, when it should be in the datetime timezone unspecified format (secretly UK datetime).

Fundamentally, datetimes passed to the backend and parsed as a C# DateTime should be in the EXACT same format as the database string format: i.e "from": "2025-07-01T09:40:00".

We don't specify in the API that datetime objects are UK or do any conversions, but so long as we always pass things around consistently it should be okay. The backend code: 
```
using (metricsRecorder.BeginScope("GetBookingsInDateRange"))
        {
            return await bookingStore.RunQueryAsync<Booking>(b => b.DocumentType == "booking" && b.Site == site && b.From >= from && b.From <= to);
        }
```
needs to have the DB datetime and the Query datetime in the same format to work. I am not suggesting any changes to the backend for now 👍 

Also fixed 'endOf(day)' as is more in line with a previous fix, and removed the fetchAvailability method as it isnt used by frontend.

**No new tests need to be written**, as my previous PRs tests catch this, its only when the API is in a different timezone does this become an issue. I have reran all the combination tests I ran before (node server in multiple different TZs, browser in multiple different TZs, and NOW the API in multiple different TZs!)

PS - I imagine this COULD have eventually had an impact to LIVE MYA due to UK/UTC differences, in the case of sessions/bookings being near midnight (1 hour shift crossing day threshold), but I suspect we dont have the data combination in it at the moment where this is an issue.

### DIFF
--- a/src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx
+++ b/src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx
@@ -2,7 +2,12 @@ import { Pagination, Spinner } from '@components/nhsuk-frontend';
 import { Site } from '@types';
 import { Suspense } from 'react';
 import { DayCardList } from './day-card-list';
-import { addToUkDatetime, DayJsType, dateFormat } from '@services/timeService';
+import {
+  addToUkDatetime,
+  DayJsType,
+  dateFormat,
+  dateTimeFormat,
+} from '@services/timeService';
 
 type Props = {
   ukWeekStart: DayJsType;
@@ -56,10 +61,7 @@ export const ViewWeekAvailabilityPage = async ({
   return (
     <>
       <Pagination previous={previous} next={next} />
-      <Suspense
-        key={ukWeekStart.format('YYYY-MM-DDTHH:mm:ssZZ')}
-        fallback={<Spinner />}
-      >
+      <Suspense key={ukWeekStart.format(dateTimeFormat)} fallback={<Spinner />}>
         <DayCardList
           site={site}
           ukWeekStart={ukWeekStart}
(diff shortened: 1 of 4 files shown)


src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx
```diff
@@ -56,10 +61,7 @@ export const ViewWeekAvailabilityPage = async ({
   return (
     <>
       <Pagination previous={previous} next={next} />
-      <Suspense
-        key={ukWeekStart.format('YYYY-MM-DDTHH:mm:ssZZ')}
-        fallback={<Spinner />}
-      >
+      <Suspense key={ukWeekStart.format(dateTimeFormat)} fallback={<Spinner />}>
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx:64]: I saw this on the first PR and almost raised it, but I assumed you kept the string in this case because it's got an extra level of precision

(though actually now I think about it, that extra precision is absolutely unnecessary so yeah all fine)

Comment (jsed-nhs) [reply src/client/src/app/site/[site]/view-availability/week/view-week-availability-page.tsx:64]: Yea just noticed it in this PR :D

sambiramairelogic (approved):

PR #650, opened 2025-04-24
APPT-662 - Don't show provisional bookings on the count

ste-banks:
Also adding extra seed data with provisional bookings to ensure e2e tests still pass for the count of booked / unbooked appointments


data/CosmosDbSeeder/items/local/booking_data/seed_booking_29.json
sambiramairelogic, line 1: Could we also add the matching booking_index files for these? 

I know they're not strictly needed for the query bookings call but I'd like to keep the seed data as accurate a reflection of real data as possible

ste-banks: Updated

data/CosmosDbSeeder/items/local/index_data/seed_booking_29_index.json
sambiramairelogic, line 1: Sorry to be a pain, but could we just update the `reference`, `status` and `id` fields so they match their respective bookings too? (looks like they all share the same `from` / `created` so you can get away with leaving those)

jsed-nhs (changes requested):
Argument to be made to filter the provisionals out of the fetchBookings API call altogether? Is there any use of returning these back, I suspect the two places its currently used it will filter out provisionals, so why pass them across the wire? Interestingly, the dailyAppointments page DOESNT include provisionals (which is why you've not made the change there...?) - just interesting regarding our debate over whether to keep provisionals in or not, they arent represented in the 'scheduled' section, but ARE included in the 'orphaned' section... very confusing!

sambiramairelogic (dismissed):

sambiramairelogic (approved):

jsed-nhs (approved):

PR #655, opened 2025-04-30
APPT-853 & APPT-839 Availability and Booking Services refactoring and Source Of Truth for Allocation

jsed-nhs:
ServiceRefactor

### Refactor: CQRS Split for Availability and Booking Services + BookingAvailabilityStateService Extraction
Summary of Changes:

- Split services by responsibility (CQRS pattern):
  - AvailabilityService → AvailabilityWriteService (AWS) and AvailabilityQueryService (AQS)
  - GetSlots moved to AQS.
  - BookingService → BookingWriteService (BWS) and BookingQueryService (BQS)
  - MakeBooking, CancelBooking, RecalculateAppointmentStatuses moved to BWS (previously in AvailabilityService to work around circular deps).
  - GetBooking* methods moved to BQS.

- New BookingAvailabilityStateService:
   - Centralised logic for greedy/BFM allocation.
   - Serves as the single source of truth for slot allocation across all MYA code paths (used by NBS queries, booking, deleting availability, etc).

- Feature toggle usage reduced from 7 → 4.

- Legacy support:
   - Old code paths maintained
   - AvailabilityCalculator retained and marked [Obsolete] for eventual removal when MultipleServices flag is deprecated.

### Test Coverage:

_Unit tests:_

- Existing tests split per new service structure (e.g. BookingServiceTests → BookingWriteServiceTests and BookingQueryServiceTests).
- Test for QueryAvailabilityFunction validates correct pathing based on MultipleServices flag.

_Integration (Gherkin) tests:_

- All code paths using the MultipleServices toggle are tested in both enabled/disabled states.
- Tests run serially to avoid race conditions with flag state.

_E2E tests:_

- Availability-related E2E tests combined into Availability.spec to wrap them under the toggle state
- Ensures current E2E suite passes for both toggle states.
- Will require extension for new frontend flows in follow-up PRs.

**Follow-up Work:**

- Add more MultipleServices-enabled tests for the new greedy allocation model and its edge cases.
- Introduce GetBookingsForSlot method to improve precision during state building.
- Integrate frontend with BookingAvailabilityStateService when MultipleServices is enabled.


features/dev.feature.flags.json
```diff
@@ -1,11 +1,10 @@
 {
   "FeatureManagement": {
     "OktaEnabled": true,
-    "MultiServiceAvailabilityCalculations": false,
+    "MultipleServicesEnabled": false,
```
sambiramairelogic, line 4: Just a minor point but can we avoid putting `Enabled` at the end of feature toggle names? I think it's redundant as a toggle by its very nature is enabled/disabled. Also you end up writing "if MyFeatureEnabled.Enabled == true" :D 

It's been bugging me that the Okta flag is named that way too

src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
```diff
@@ -2,39 +2,45 @@ namespace Nhs.Appointments.Core;
...
     {
-        var allSessions = await availabilityStore.GetSessions(site, from, until);
+        var allSessions = await availabilityStore.GetSessions(site, from, until, "*");
         var filteredSessions = service == "*"
             ? allSessions
             : allSessions.Where(s => s.Services.Contains(service));
 
         var bookings = await bookingDocumentStore.GetInDateRangeAsync(from.ToDateTime(new TimeOnly(0, 0)),
-            until.ToDateTime(new TimeOnly(23, 59)), site);
+            until.ToDateTime(new TimeOnly(23, 59)), site, "*");
```
sambiramairelogic, line 29: This method is marked obsolete but there's a logical change here. I know `GetInDateRangeAsync` behaves the same was as it did before when provided an asterisk and this would be absolutely fine, but technically this is a change.

 I think we need to decide how pedantic we're gonna be - atm (in main) we're being super strict about not changing anything at all, but obviously that's generated mess. I'm happy to make changes like this but I would just say if we're making any change at all we might as well go the whole hog - 
 
 Just as a general attitude in this refactor I say let's either avoid any changes like this or do them properly and clear out all the mess

src/api/Nhs.Appointments.Core/AllocationState.cs
sambiramairelogic, line 1: I don't know how I feel about the name `AllocationState` yet - it feels a bit misleading since we're not actually allocating bookings. 

What's actually on this object is: 
* The bookings within that time frame whose status remains `Booked`
* The availability for additional bookings that remains, considering the above bookings exists

Neither one of these indicates which session a booking has been "allocated" to, nor would that be true information if it did. I actually think `AvailabilityState` conveys this more clearly, but I presume you renamed it for a reason so happy to come up with something else. Perhaps some sort of `RemainingAvailabilityState`, or `SiteCalendarState` I don't know - just something that indicates it's giving you remaining availability

src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
jsed-nhs: Agreed I probably should have made this as a new method with the service overload, rather than changing the current one, so as not to touch 'old' code. I am happy to do this.

src/api/Nhs.Appointments.Core/AllocationState.cs
jsed-nhs: True, I just wanted a name to distinguish itself clearly from the AvailabilityServices. Happy to rename to something else, I chose Allocation because the code is the algorithm for how we allocate bookings to slots sitewide.

sambiramairelogic: Other contenders - `BookingAvailabilitySnapshot`, `BookingAvailabilityState`, or just simply `ScheduledBookingsAndAvailability`

features/dev.feature.flags.json
jsed-nhs: Agreed

src/api/Nhs.Appointments.Api/Functions/QueryAvailabilityFunction.cs
```diff
@@ -74,7 +75,21 @@ await Parallel.ForEachAsync(request.Sites, async (site, ct) =>
 
     private async Task<QueryAvailabilityResponseItem> GetAvailability(string site, string service, QueryType queryType, DateOnly from, DateOnly until, int consecutive)
     {
-        var slots = (await availabilityCalculator.CalculateAvailability(site, service, from, until)).ToList();
+        var dayStart = from.ToDateTime(new TimeOnly(0, 0));
```
pata9, line 78: No need to put these variables here, also wondering why we need a datetime for the new logic, is this the requirement for time creeping in for query availability?

src/api/Nhs.Appointments.Core/AllocationStateService.cs
```diff
@@ -0,0 +1,65 @@
+namespace Nhs.Appointments.Core;
+
+public class AllocationStateService(
+    IAvailabilityQueryService availabilityQueryService,
+    IBookingQueryService bookingQueryService) : IAllocationStateService
+{
+    public async Task<AllocationState> Build(string site, DateTime from, DateTime to, string service, bool processRecalculations = true)
```
pata9, line 7: I think we can remove the concept of `processRecalculations` from this service, I don't think it needs to be in here. We should be able to return the state of play here and then do this after calculating the state. There's a few ways we can do this

src/api/Nhs.Appointments.Core/BookingWriteService.cs
```diff
@@ -0,0 +1,345 @@
...
+public interface IBookingWriteService
+{
+    Task<(bool Success, string Reference)> MakeBooking(Booking booking);
+
+    Task<BookingCancellationResult> CancelBooking(string bookingReference, string site);
+
+    Task<bool> SetBookingStatus(string bookingReference, AppointmentStatus status,
+        AvailabilityStatus availabilityStatus);
+
+    Task SendBookingReminders();
```
pata9, line 17: We should raise some tech debt here `SendBookingReminders` is too business logicy :D I reckon this method is the only reason we have gov notify as a dependency here?

Reckon we should aim to have `GetRemainingReminders` in the read and `SetReminderStatus` in here.

src/api/Nhs.Appointments.Core/BookingWriteService.cs
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        return await MakeBooking_SingleService(booking);
+    }
+
+    public async Task<BookingCancellationResult> CancelBooking(string bookingReference, string site)
+    {
+        if (await featureToggleHelper.IsFeatureEnabled(Flags.MultipleServicesEnabled))
+        {
+            return await CancelBooking_MultipleServices(bookingReference, site);
```
pata9, line 57: The longer we maintain this separation/duplication the more risk we're adding here, I'll have a think but I think I put in a PR to the branch for you guys to review

src/api/Nhs.Appointments.Core/IAvailabilityStore.cs
```diff
@@ -2,7 +2,7 @@ namespace Nhs.Appointments.Core;
 
 public interface IAvailabilityStore
 {
-    Task<IEnumerable<SessionInstance>> GetSessions(string site, DateOnly notBefore, DateOnly notAfter);
+    Task<IEnumerable<SessionInstance>> GetSessions(string site, DateOnly from, DateOnly to, string service);
```
pata9, line 5: Why's service been added here? No matter what we want to also be recalculation the whole state don't we? I'll track down the uses here and see if I can understand the reasoning!

src/api/Nhs.Appointments.Core/IBookingsDocumentStore.cs
```diff
@@ -5,7 +5,7 @@ namespace Nhs.Appointments.Core;
 public interface IBookingsDocumentStore 
 {
     Task InsertAsync(Booking booking);
-    Task<IEnumerable<Booking>> GetInDateRangeAsync(DateTime from, DateTime to, string site);
+    Task<IEnumerable<Booking>> GetInDateRangeAsync(DateTime from, DateTime to, string site, string service);
```
pata9, line 8: Again same for the availability, why are we filtering on service?

src/api/Nhs.Appointments.Core/AllocationStateService.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        var allocationState = new AllocationState();
+
+        var orderedLiveBookings = await bookingQueryService.GetOrderedLiveBookings(site, from, to, service);
+        var slots = await availabilityQueryService.GetSlots(site, DateOnly.FromDateTime(from), DateOnly.FromDateTime(to), service);
```
pata9, line 12: Ye I need to understand why we're filtering on service here, Sessions without the queried service can impact Availability/Allocation for a service, we shouldn't be filtering here?

src/api/Nhs.Appointments.Core/IAvailabilityStore.cs
pata9: PR to cover this here - https://github.com/NHSDigital/nbs-appointments-management-service/pull/660

src/api/Nhs.Appointments.Core/IBookingsDocumentStore.cs
pata9: PR to cover this here - https://github.com/NHSDigital/nbs-appointments-management-service/pull/660

src/api/Nhs.Appointments.Core/AllocationStateService.cs
jsed-nhs: I think the idea is that figuring out what needs recalculating is intrinsically tied to the Greedy/BFM allocation algorithm while in progress. I can't off top of my head think how you would deduce this after the state has been calculated? I agree it currently looks crap and I'm not happy with how it reads, but I didn't want to create two methods doing the same logic :(  lets see if we can come up with a cleaner way of achieving this.

src/api/Nhs.Appointments.Core/BookingWriteService.cs
jsed-nhs: We need the separation and duplication while we support both current Live imp. and new multi services imp. This would/SHOULD be all cleaned up when the toggles are removed and tidied. This is a downside to FT development, you end up maintaining/duplicating old code and having to manage both of it, and encourage it to be cleaned up at some point.

src/api/Nhs.Appointments.Core/AllocationState.cs
jsed-nhs: AllocationSnapshotService? I don't know why I like the word Allocation a lot for this logic....

src/api/Nhs.Appointments.Core/BookingWriteService.cs
jsed-nhs: Agreed, refactor another day

jsed-nhs:
Confirmed today that all the existing E2E tests pass when ran manually with MultiServices enabled/disabled. Now to add new targeted MultiService tests...

tests/Nhs.Appointments.Api.UnitTests/Functions/QueryAvailabilityFunctionTests.cs
```diff
@@ -218,6 +232,87 @@ public async Task RunAsync_CallsGrouperWithCorrectSlots_ForEachDayInRequest()
...
+            QueryType.Days, 
+            null);
+
+        var httpRequest = CreateRequest(request);
+
+        await _sut.RunAsync(httpRequest);
+        
+        _availabilityCalculator.Verify(x => x.CalculateAvailability("2de5bb57-060f-4cb5-b14d-16587d0c2e8f", "COVID", new DateOnly(2077, 01, 01),
+                new DateOnly(2077, 01, 03)),
+            Times.Exactly(0));
```
sambiramairelogic, line 311: Not a comment as this is equivalent but just to let you know you can do `Times.Never` and `Times.Once` here instead of using Exactly :)

tests/Nhs.Appointments.Core.UnitTests/AllocationState/AllocationStateServiceTestBase.cs
```diff
@@ -1,32 +1,19 @@
...
 
-namespace Nhs.Appointments.Core.UnitTests.AvailabilityCalculations;
-
-[MockedFeatureToggle("MultiServiceAvailabilityCalculations", true)]
-public class AvailabilityCalculationsBase : FeatureToggledTests
+public class AllocationStateServiceTestBase
 {
-    protected readonly AvailabilityService _sut;
+    protected readonly AllocationStateService Sut;
+    protected readonly Mock<TimeProvider> TimeProvider = new();
```
sambiramairelogic, line 6: Doesn't really matter but you've changed the casing here - old values were consistent with `_foo` but new ones are `Foo`

tests/Nhs.Appointments.Core.UnitTests/AllocationState/MultiServiceTests.cs
```diff
@@ -25,19 +25,19 @@ public async Task MultipleServicesByCreatedDate()
         };
 
         SetupAvailabilityAndBookings(bookings, sessions);
-
-        var resultingAvailabilityState = await _sut.GetAvailabilityState(MockSite, new DateOnly(2025, 1, 1));
+            
+        var resultingAllocationState = await Sut.BuildRecalculations(MockSite, new DateTime(2025, 1, 1, 9, 0, 0), new DateTime(2025, 1, 1, 9, 10, 0));
```
sambiramairelogic, line 29: Likewise here (real important comment, I know... ;P )

src/api/Nhs.Appointments.Core/AllocationStateService.cs
```diff
@@ -0,0 +1,108 @@
...
+    public static void AddNoLongerSupportedBookings(this List<BookingAvailabilityUpdate> recalculations, AllocationState state, List<Booking> bookings)
+    {
+        recalculations.AddRange(bookings
+            .Where(booking => !state.SupportedBookingReferences.Contains(booking.Reference) &&
+                              booking.AvailabilityStatus is AvailabilityStatus.Supported &&
+                              booking.Status is not AppointmentStatus.Provisional)
+            .Select(booking => new BookingAvailabilityUpdate(booking, AvailabilityUpdateAction.SetToOrphaned)));
+    }
+
+    public static void DeleteRemainingUnsupportedProvisionalBookings(this List<BookingAvailabilityUpdate> recalculations, AllocationState state, List<Booking> bookings)
```
sambiramairelogic, line 101: This method is named differently from the others, which are `AddFoo...`. The name difference made me assume this method was actually doing the delete

src/api/Nhs.Appointments.Core/AllocationStateService.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    public static void AddNewlySupportedBookings(this List<BookingAvailabilityUpdate> recalculations, AllocationState state, List<Booking> bookings)
+    {
+        recalculations.AddRange(bookings
+            .Where(booking => state.SupportedBookingReferences.Contains(booking.Reference) &&
+                              booking.AvailabilityStatus is not AvailabilityStatus.Supported)
+            .Select(booking => new BookingAvailabilityUpdate(booking, AvailabilityUpdateAction.SetToSupported)));
+    }
+
+    public static void AddNoLongerSupportedBookings(this List<BookingAvailabilityUpdate> recalculations, AllocationState state, List<Booking> bookings)
```
sambiramairelogic, line 92: Each one of these methods loops over the `bookings` object, which could be potentially very large. Would it not be better to loop over it once instead of 3 times?

src/api/Nhs.Appointments.Core/AllocationStateService.cs
```diff
@@ -0,0 +1,108 @@
...
+            .FirstOrDefault();
+
+    private async Task<IEnumerable<Booking>> GetBookings(string site, DateTime from, DateTime to) =>
+        await bookingQueryService.GetOrderedLiveBookings(site, from, to);
+
+    private async Task<IEnumerable<SessionInstance>> GetSlots(string site, DateTime from, DateTime to) =>
+        await availabilityQueryService.GetSlots(site, DateOnly.FromDateTime(from), DateOnly.FromDateTime(to));
+}
+
+public static class RecalculationExtensions
```
sambiramairelogic, line 82: Each one of these methods loops over the bookings object, which could be potentially very large. Would it not be better to loop over it once instead of 3 times?

src/api/Nhs.Appointments.Core/AllocationStateService.cs
```diff
@@ -0,0 +1,108 @@
...
+    IAvailabilityQueryService availabilityQueryService,
+    IBookingQueryService bookingQueryService) : IAllocationStateService
+{
+    public async Task<AllocationState> BuildAllocation(string site, DateTime from, DateTime to)
+    {
+        var (bookings, slots) = await FetchData(site, from, to);
+        return BuildAllocation(bookings, slots);
+    }
+
+    public async Task<IEnumerable<BookingAvailabilityUpdate>> BuildRecalculations(string site, DateTime from,
```
sambiramairelogic, line 13: What's the motivation for splitting this out from `BuildAllocation`? I'm concerned a future dev might go 

"build my allocation. Go do some logic or whatever. Now build my recalculations and process..." without realising they've done an expensive calculation twice. 

An extra benefit of keeping this object on the AllocationState is the extra assurance that it's empty if we're querying

src/api/Nhs.Appointments.Core/AllocationStateService.cs
pata9: These will only be run on recalculations so I don't think we should be too concerned here, the recalculations would always be constrained to a subset... maximum one day? If that's not the case we should look to how we can improve there.

src/api/Nhs.Appointments.Core/AllocationStateService.cs
pata9: Just simply separation of concerns, they are different things used by different purposes

src/api/Nhs.Appointments.Core/AvailabilityWriteService.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        await availabilityCreatedEventStore.LogTemplateCreated(site, from, until, template, user);
+    }
+
+    public async Task ApplySingleDateSessionAsync(DateOnly date, string site, Session[] sessions,
+        ApplyAvailabilityMode mode, string user, Session sessionToEdit = null)
+    {
+        await SetAvailabilityAsync(date, site, sessions, mode, sessionToEdit);
+        await bookingWriteService.RecalculateAppointmentStatuses(site, date);
```
sambiramairelogic, line 41: I'd be tempted to move the RecalculateAppointmentStatuses call inside SetAvailabilityAsync, because future devs might call SetAvailabilityAsync and forget to do the recalculation.

src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
```diff
@@ -2,39 +2,47 @@ namespace Nhs.Appointments.Core;
...
+    public async Task<IEnumerable<SessionInstance>> CalculateAvailability(string site, string service, DateOnly from,
+        DateOnly until)
     {
         var allSessions = await availabilityStore.GetSessions(site, from, until);
         var filteredSessions = service == "*"
             ? allSessions
             : allSessions.Where(s => s.Services.Contains(service));
 
+        //TODO bookings not filtered by service in SingleService code path.
+        //what happens if we enable MultiServices then disable?
```
sambiramairelogic, line 29: The check during GetAvailableSlots decrements capacity for each slot with a matching service type, so looks like they'll still consume availability. Moreover, is it even possible to switch multiple services off once it's switched on? If there's availability and bookings for multiple services in the DB, I think we say this is a hard point of no return.

src/api/Nhs.Appointments.Core/AllocationStateService.cs
sambiramairelogic: I don't think they are different things, as one directly depends upon the other. Every time you're genuinely recalculating you need to do the recalculations, and if you're querying availability then the recalculations will be always be empty anyway so won't slow the method down at all (plus if they're not empty that tells us there's a bug!)

I think splitting these out introduces risk for no benefit. If we *really* want to be able to build state without recalculations, I'd make that an optional parameter on the method. Something like `skipRecalculations = false` that you can set to true. But like I say, I don't think there's a case for it

pata9: So are you saying that any change to this method would require a change to the other? They **need** to be coupled in the same method? I don't think I fully understand your concern here, don't see any risk in the separation

src/api/Nhs.Appointments.Core/BookingWriteService.cs
```diff
@@ -0,0 +1,343 @@
...
+        {
+#pragma warning disable CS0618 // Keep availabilityCalculator around until MultipleServicesEnabled is stable
+            var slots = (await availabilityCalculator.CalculateAvailability(booking.Site, booking.Service,
+                    DateOnly.FromDateTime(booking.From.Date), DateOnly.FromDateTime(booking.From.Date.AddDays(1))))
+                .ToList();
+#pragma warning restore CS0618 // Keep availabilityCalculator around until MultipleServicesEnabled is stable
+
+            var canBook = slots.Exists(sl => sl.From == booking.From && sl.Duration.TotalMinutes == booking.Duration);
+
+            if (canBook)
```
pata9, line 136: from this canbook check there is no difference between the single and multiple, worth moving to another method? Reduce the need to maintain both versions? Not sure how long both versions will live tbh meaning how worth while the refactor would be?

src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
jsed-nhs: Agree, it's just something I noticed during the refactor, that if we ever went back to 'MultipleServices disabled' but with multi services in the DB, it would cause big problems. Think I just need to highlight the lack of 'go back' once an environment gets enabled.

src/api/Nhs.Appointments.Core/BookingWriteService.cs
jsed-nhs: Yea agree with latter, the SingleService methods will be deleted at some point, so any refactor to be used by both will not serve useful for very long.

tests/Nhs.Appointments.Core.UnitTests/AllocationState/AllocationStateServiceTestBase.cs
jsed-nhs: My intellisense complains. Since the Sut is protected and not private, it wants the casing to be this. When the sut is private only I agree it should be _sut. Happy to revert, but my intellisense doesnt like it :D

src/api/Nhs.Appointments.Core/AllocationStateService.cs
jsed-nhs: Agree, it looks a lot neater this way, but may be more expensive. Will have a look and see if can keep it efficient but clean.

src/api/Nhs.Appointments.Core/BookingQueryService.cs
```diff
@@ -0,0 +1,55 @@
...
+    public async Task<List<Booking>> GetOrderedLiveBookings(string site, DateTime from, DateTime to)
+    {
+        var bookings = (await GetBookings(from, to, site))
+            .Where(b => _liveStatuses.Contains(b.Status))
+            .Where(b => !IsExpiredProvisional(b))
+            .OrderBy(b => b.Created)
+            .ToList();
+
+        return bookings;
+    }
```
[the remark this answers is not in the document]
jsed-nhs: LINQ is smart enough, it builds a chain of predicates, but only one iterator is created when you enumerate it with a ToList or foreach call. its essentially the same as combining them both into a single .Where but reads cleaner this way.

src/api/Nhs.Appointments.Core/AvailabilityCalculator.cs
jsed-nhs: I've just thought, I may have to address this in the next PR (for adding new Int/E2E tests for multiple services) as it will have mulliple service data in the DB etc, so the single service tests might fail. Only not an issue for now because this PR doesn't contain any 'new' MultipleService functionality tests.

src/api/Nhs.Appointments.Core/AllocationStateService.cs
jsed-nhs: I originally had the `skipRecalculations` bool in the one method as you suggested, however it never looked that clean to me, and I didn't like the idea of the Recalculations object being in the result when it was possibly Empty/Null. I liked Pauls idea of splitting them out and I personally feel it does add some value. Having said that, I agree with your concern of a dev possibly calling both methods needlessly. I suggest leaving the code as is, but adding some detailed //summary around the Interface/Methods to help devs realise all the methods in this Service go through an expensive Allocation algorithm, so to help avoid needless double calls. If a scenario ever arises where we might to do a third new operation that also relies on the AllocationState, we can just add a new method to the interface and have it be called once.

jsed-nhs: With the IEnumerable and ToList recent refactor also, the BuildAllocation method can keep the bookings as IEnum, whereas the Recalculations method requires a call to .ToList, so this is a further benefit of them being separate.

src/api/Nhs.Appointments.Core/AllocationStateService.cs
DACA16: We will cover this off and confirm via Performance tests if the algorithm is causing issues too - how about we leave this as is for Version 1, and lets evaluate the impact later

src/api/Nhs.Appointments.Core/AllocationState.cs
DACA16: Id vote for BookingAvailabilityState

DACA16: Lets just go with BookingAvailabilityState please

ste-banks: Another upvote for BookingAvailabilityState

src/api/Nhs.Appointments.Core/BookingWriteService.cs
```diff
@@ -0,0 +1,343 @@
...
+    public Task<IEnumerable<string>> RemoveUnconfirmedProvisionalBookings()
+    {
+        return bookingDocumentStore.RemoveUnconfirmedProvisionalBookings();
+    }
+
+    public async Task RecalculateAppointmentStatuses(string site, DateOnly day)
+    {
+        if (await featureToggleHelper.IsFeatureEnabled(Flags.MultipleServices))
+        {
+            await RecalculateAppointmentStatuses_MultipleServices(site, day);
```
ste-banks, line 118: Should we be returning here because at the moment this recalculate statuses for both multiple services & single service if the feature flag is enabled?

jsed-nhs: Great spot, will fix and then make sure this bug of it running both times is tested for in the follow-up PR for MultipleServicesEnabled tests.

src/api/Nhs.Appointments.Core/AllocationStateService.cs
sambiramairelogic: > These will only be run on recalculations so I don't think we should be too concerned here, the recalculations would always be constrained to a subset... maximum one day? If that's not the case we should look to how we can improve there.

I don't we can rely on the recalculations being small - someone could apply an availability template over a whole month (or even a whole year). Equally whilst it's currently only possible to cancel a single session, it's in the plan to let people cancel multiple sessions in a single request so potentially cancelling a whole month/year too.

sambiramairelogic: Splitting it out into multiple processes doesn't reduce the number of calculations in any way - 
1 process doing 1000 calculations is 1000
10 processes doing 100 calculations each is still 1000

src/api/Nhs.Appointments.Core/AllocationState.cs
jsed-nhs: BookingAvailabilityState decided

tests/Nhs.Appointments.Core.UnitTests/AllocationState/AllocationStateServiceTestBase.cs
jsed-nhs: Leave as is for now

tests/Nhs.Appointments.Core.UnitTests/AllocationState/MultiServiceTests.cs
jsed-nhs: Leave as is for now

src/api/Nhs.Appointments.Core/BookingWriteService.cs
jsed-nhs: Will leave for now.

src/api/Nhs.Appointments.Core/BookingWriteService.cs
jsed-nhs: The noticed Live issue regarding SingleService code path using expired provisionals in its Recalculation logic - to raise bug with impact, TBD fixed or not. Should be fixed and tested for Multiple Services code path.

jsed-nhs:
Future PR: TimePeriod refactor to replace Duration (timespan) with an int DurationMins.

ste-banks (approved):

PR #660, opened 2025-05-06
remove services filter from the availability calculations

pata9:
As a rule of thumb we should never be calculating availability without considering the entire state for that slot. This PR removes any filtering on service for the new interfaces

### DIFF
--- a/src/api/Nhs.Appointments.Core/BookingWriteService.cs
+++ b/src/api/Nhs.Appointments.Core/BookingWriteService.cs
@@ -161,10 +161,10 @@ public class BookingWriteService(
             var from = booking.From;
             var to = booking.From.AddMinutes(booking.Duration);
 
-            var slots = (await allocationStateService.Build(booking.Site, from, to, booking.Service, false))
+            var slots = (await allocationStateService.Build(booking.Site, from, to, false))
                 .AvailableSlots;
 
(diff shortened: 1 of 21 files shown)


src/api/Nhs.Appointments.Core/BookingWriteService.cs
```diff
@@ -161,10 +161,10 @@ public async Task RecalculateAppointmentStatuses(string site, DateOnly day)
             var from = booking.From;
             var to = booking.From.AddMinutes(booking.Duration);
 
-            var slots = (await allocationStateService.Build(booking.Site, from, to, booking.Service, false))
+            var slots = (await allocationStateService.Build(booking.Site, from, to, false))
                 .AvailableSlots;
 
-            var canBook = slots.Any(sl => sl.From == booking.From && sl.Duration.TotalMinutes == booking.Duration);
+            var canBook = slots.Any(sl => sl.Services.Contains(booking.Service) && sl.From == booking.From && sl.Duration.TotalMinutes == booking.Duration);
```
jsed-nhs, line 167: I think this canBook logic was correct and needs reverting, it isn't related to the service-in-services changes I made.

jsed-nhs (changes requested):
I agree in principle with needing all services when building allocation state. However I remember doing this change for a reason that I can't remember right now! Need to chat about it when I'm in this afternoon.

jsed-nhs (approved):

PR #661, opened 2025-05-06
APPT-704: Okta auth route

sambiramairelogic:
This PR creates codepaths for creating, editing and deleting users both with and without the Okta feature flag. The key motivator for this work was that the UI has changed substantially from its original design (new designs in the [APPT-704 ticket](https://nhsd-jira.digital.nhs.uk/browse/APPT-704)), and collecting the user's first/last names is now only required if the User is A) an Okta user and B) does not yet exist in Okta. 

Key changes: 
- The Okta code has been brought into the Core project because Core needs to reference it directly
- There is a new API which, for a given email address, will inform on whether that address already exists in MYA and its relevant Identity Provider
- There is a new store for email domain whitelists which is sensitive to the Okta feature toggle
- The new UI route has been created
- Unit, Integration and E2E tests for all the above

This already a MAHOOSIVE pull request, so there are some additional changes I'll do as further tickets: 
- I'd like much more substantial testing of these features with the toggle on/off in Playwright. For now I've simply updated the existing tests to match the new UI. I'd like to implement this using the tags feature so we can selectively run tests in different modes, but that'll trigger a fair raft of refactors to existing tests. 
- The notification triggered after creating/editing a user is going to be re-implemented in [its own ticket](https://nhsd-jira.digital.nhs.uk/browse/APPT-799)


src/api/Nhs.Appointments.Api/Functions/ProposePotentialUserFunction.cs
```diff
@@ -0,0 +1,47 @@
...
+) : BaseApiFunction<ProposePotentialUserRequest, ProposePotentialUserResponse>(validator, userContextProvider, logger,
+    metricsRecorder)
+{
+    [OpenApiOperation("ProposePotentialUser", "user/propose-potential",
+        Summary = "Gets the current status of a potential new user. Used to inform the user creation process.")]
+    [OpenApiResponseWithBody(HttpStatusCode.OK, "application/json", typeof(ProposePotentialUserResponse),
+        Description = "The response containing the current status of a potential new user.")]
+    [Function("ProposePotentialUserFunction")]
+    public override Task<IActionResult> RunAsync(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "user/propose-potential")]
```
pata9, line 29: Should this be a post? Feels like a get to me? Unless `GetUserIdentityStatusAsync` does something then that method needs renaming. Also do we need the SiteId here?

src/api/Nhs.Appointments.Core/UserService.cs
```diff
@@ -89,6 +96,51 @@ public Task<OperationResult> RemoveUserAsync(string userId, string site)
...
+    private async Task<bool> CheckIfUserExistsAtSite(string siteId, string userId)
+    {
+        var userProfile = await userStore.GetOrDefaultAsync(userId);
+
+        if (userProfile is null)
+        {
+            return false;
+        }
+
+        return userProfile.RoleAssignments.Any(roleAssignment =>
```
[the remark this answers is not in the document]
pata9: Could just return that the user exists instead of checking the scope?

The global check confuses me as well?

src/api/Nhs.Appointments.Persistance/Models/WhitelistedEmailDomainsDocument.cs
```diff
@@ -0,0 +1,10 @@
+using Newtonsoft.Json;
+
+namespace Nhs.Appointments.Persistance.Models;
+
+[CosmosDocumentType("system")]
+public class WhitelistedEmailDomainsDocument : CoreDataCosmosDocument
+{
+    [JsonProperty("domains")]
+    public IEnumerable<string> Domains { get; set; }
```
pata9, line 9: Annoyingly it isn't all domains.. but I can't think of anything better :D

src/client/src/app/lib/services/appointmentsService.ts
```diff
@@ -275,7 +291,7 @@ export const saveUserRoleAssignments = async (
   firstName: string,
   lastName: string,
   roles: string[],
-  isEdit: boolean,
+  // isEdit: boolean,
```
pata9, line 294: guess this needs deleting

src/client/src/app/lib/services/appointmentsService.ts
```diff
@@ -284,18 +300,20 @@ export const saveUserRoleAssignments = async (
...
     JSON.stringify(payload),
   );
   handleEmptyResponse(response);
 
-  const notificationType = 'ams-notification';
-  const notificationMessage = isEdit
-    ? `You have changed a user's role.`
-    : `You have added a new user to MYA The user will be sent information about how to login.`;
-  raiseNotification(notificationType, notificationMessage);
+  // const notificationType = 'ams-notification';
```
pata9, line 310: again can we delete comments?

src/client/src/app/site/[site]/users/manage/set-user-roles-wizard.tsx
```diff
@@ -0,0 +1,107 @@
...
+          }
+        : undefined,
+    },
+    resolver: yupResolver(setUserRolesFormSchema),
+  });
+  const router = useRouter();
+
+  const userIdentityStatus = methods.watch('userIdentityStatus');
+  const isCreatingNewOktaUser =
+    userIdentityStatus?.identityProvider === 'Okta' &&
```
Comment (pata9) [inline src/client/src/app/site/[site]/users/manage/set-user-roles-wizard.tsx:50]: Should look into adding a const for identityProviders?

src/api/Nhs.Appointments.Api/Functions/ProposePotentialUserFunction.cs
sambiramairelogic: We have lots of APIs implemented as posts because we want to accept a request body. Examples include the QueryBookingsFunction and QueryAvailabilityFunction. I could make this a get but it'd need to accept 2 parameters which personally I think is messier, so chose a post. 

We do need the site id because we need to know whether this user exists at a specific site, not just whether they exist at all :)

pata9: How come we need to know there at the site? If this is the case we should probs remove the global scope check I think :/

src/api/Nhs.Appointments.Core/UserIdentityStatus.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Newtonsoft.Json;
+
+namespace Nhs.Appointments.Core;
+
+public class UserIdentityStatus
+{
+    [JsonProperty("extantInMya")] public required bool ExtantInMya { get; set; }
```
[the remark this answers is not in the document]
pata9: I agree! I only commented because it's different to the others in this file

src/client/src/app/lib/services/appointmentsService.ts
sambiramairelogic: Coming in https://nhsd-jira.digital.nhs.uk/browse/APPT-799

src/client/src/app/lib/services/appointmentsService.ts
sambiramairelogic: Coming in https://nhsd-jira.digital.nhs.uk/browse/APPT-799

src/api/Nhs.Appointments.Core/UserService.cs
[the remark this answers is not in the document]
pata9: I don't think the this is the case? If the user exists then we just add role assignments to that user don't we? Whether they're at the site or not!

src/client/src/app/site/[site]/users/manage/set-user-roles-wizard.tsx
Comment (sambiramairelogic) [reply src/client/src/app/site/[site]/users/manage/set-user-roles-wizard.tsx:50]: Already done 
image

src/client/src/app/lib/services/appointmentsService.ts
sambiramairelogic: Yep good spot

src/client/src/app/site/[site]/users/manage/set-user-roles-wizard.tsx
Comment (pata9) [reply src/client/src/app/site/[site]/users/manage/set-user-roles-wizard.tsx:50]: Cool!

src/api/Nhs.Appointments.Persistance/Models/WhitelistedEmailDomainsDocument.cs
sambiramairelogic: Isn't it? I can only see domains in the whitelist documents 

Admittedly some are only top-level domains and others include second-level, but they're all domains I believe?

src/api/Nhs.Appointments.Core/UserIdentityStatus.cs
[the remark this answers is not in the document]
sambiramairelogic: Yeah that'll be because one of the rules is about maximum line length, so parameters with longer/shorter names can differ in how they're formatted. Can change if you really strongly enough but as I say personally would rather do a big-bang formatting ticket at some point :)

src/api/Nhs.Appointments.Api/Functions/ProposePotentialUserFunction.cs
sambiramairelogic: Because if they already have roles at the site then the role selection checkboxes should be pre-populated with their existing roles. If they're a new users they should instead all default to unchecked

src/api/Nhs.Appointments.Persistance/Models/WhitelistedEmailDomainsDocument.cs
pata9: Depends how you define it :p sub include the full domain, others just the end! I think it just annoys me more than anything else :D

src/api/Nhs.Appointments.Api/Functions/ProposePotentialUserFunction.cs
pata9: Ahh okay cool, we have a driver! Understood

src/api/Nhs.Appointments.Core/UserService.cs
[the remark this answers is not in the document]
pata9: Should global pre-populate the permissions? Wonky land :D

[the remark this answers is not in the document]
sambiramairelogic: Okay have removed for now, but we will need to revisit this after the discussions about region/icb roles. If those users should appear in the UI or be editable in any way we'll need to add this back in but I agree that's a considering for future work

src/api/Nhs.Appointments.Api/Functions/ProposePotentialUserFunction.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    protected override async Task<ApiResult<ProposePotentialUserResponse>> HandleRequest(
+        ProposePotentialUserRequest request,
+        ILogger logger)
+    {
+        var userIdentityStatus = await userService.GetUserIdentityStatusAsync(request.SiteId, request.UserId);
+
+        return ApiResult<ProposePotentialUserResponse>.Success(new ProposePotentialUserResponse
+        {
+            ExtantInMya = userIdentityStatus.ExtantInMya,
```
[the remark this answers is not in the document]
sambiramairelogic: I'm perfectly happy with `ExtantInSite` - have renamed

pata9 (approved):

ste-banks (approved):

PR #670, opened 2025-05-09
APPT-803 - seed clinical services for int

pata9:


jsed-nhs (changes requested):
Should we be putting the TestA/B services into Int? wouldn't it be nicer to put some real service names in so we aren't polluting the environment?

pata9:
> Should we be putting the TestA/B services into Int? wouldn't it be nicer to put some real service names in so we aren't polluting the environment?

This is just for the testers benefit, this will be overridden by future real services tickets. I'm deliberately not trying to be too close to real data

jsed-nhs (approved):

PR #672, opened 2025-05-09
APPT-855/terraform scripts

pata9:
This PR adds a retention period on to all builds that go to staging and above and refactors existing terraform actions into scripts

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -347,5 +347,5 @@ stages:
           cosmosAccountName: nbs-mya-cdb-${{env}}-uks
           resourceGroup: nbs-mya-rg-${{env}}-uks
           serviceConnectionName: nbs-mya-rg-${{env}}
-          isMain: true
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -325,5 +333,5 @@ stages:
           cosmosAccountName: nbs-mya-cdb-${{env}}-uks
           resourceGroup: nbs-mya-rg-${{env}}-uks
           serviceConnectionName: nbs-mya-rg-${{env}}
-          isMain: ${{eq(variables['Build.SourceBranch'], 'refs/heads/main')}}
+          isMain: true
```
pata9, line 336: need to revert this before merge

pata9:
> lgtm, but I'd like to run the disaster recovery process with these changes before approving. Should be free this aft if you wanna pair up for that

We're in a bit of a pickle there, as we need to release to staging to test the DR :D, I think this is a risk we'll have to take. Just need to communicate it!

PR #674, opened 2025-05-12
APPT-877 - AW clinical services added to mya

pata9:
Clinical Services planned for Autumn/Winter added into MYA. Requires MultipleServices feature flag


data/CosmosDbSeeder/items/int/core_data/clinical_services.json
```diff
@@ -7,40 +7,44 @@
             "label": "RSV (Adult)"
```
sambiramairelogic, line 7: This differs from the text in the ticket

pata9: See here - https://nhsuk.slack.com/archives/C062RBW4NF4/p1747131185481079

Good to go!

pata9: Wait Lauren wants to wait!

pata9: Should be good to go now!

PR #675, opened 2025-05-13
APPT-880 - Enabling the high load function for Bulk Import

ste-banks:
We hit an issue on int where uploading 244 Okta users caused the BulkImportFunction to hit it's default timeout (5 minutes) so for staging / production we want to enable the high load function for this which has a longer timeout (30 minutes as default).

Once we've figured out a limit on staging for high load, I will work with Anna to workout a plan over a number of weeks to upload all users (both nhs & Okta) so that we don't hit the timeout.

### DIFF
--- a/infrastructure/resources/http_functions.tf
+++ b/infrastructure/resources/http_functions.tf
@@ -78,6 +78,7 @@ resource "azurerm_windows_function_app" "nbs_mya_http_func_app" {
     "AzureWebJobs.RemoveUnconfirmedProvisionalBookings.Disabled"   = true
     "AzureWebJobs.ClearLocalFeatureFlagOverridesFunction.Disabled" = true
     "AzureWebJobs.SetLocalFeatureFlagOverrideFunction.Disabled"    = true
+    "AzureWebJobs.BulkImportFunction.Disabled"                     = var.disable_bulk_import_function
   }
 
   sticky_settings {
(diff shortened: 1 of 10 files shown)


sambiramairelogic (changes requested):
I think we need to disable this in http_functions as well as enabling it in high load, otherwise we might still hit the http one

ste-banks:
> I think we need to disable this in http_functions as well as enabling it in high load, otherwise we might still hit the http one

Yeah good spot, I'll update that.

infrastructure/resources/variables.tf
```diff
@@ -272,3 +272,7 @@ variable "app_config_connection" {
   default = ""
   sensitive = true
 }
+
+variable "disable_bulk_import_function" {
```
sambiramairelogic, line 276: (not a blocking comment, happy with this if you disagree)

Could I suggest that instead of creating this new variable we instead derive the enabled/state state of the bulk upload import by the value of `create_high_load_function_app` which already exists? I think this would be slightly clearer too, as the current naming of this new variable isn't as obvious that it's being enabled elsewhere

ste-banks: I went for consistency with the query availability variable for this as it's the same logic. That has it's own variable `disable_query_availability_function` instead as well.

infrastructure/resources/http_functions.tf
```diff
@@ -160,6 +161,7 @@ resource "azurerm_windows_function_app_slot" "nbs_mya_http_func_app_preview" {
     "AzureWebJobs.RemoveUnconfirmedProvisionalBookings.Disabled"   = true
     "AzureWebJobs.ClearLocalFeatureFlagOverridesFunction.Disabled" = true
     "AzureWebJobs.SetLocalFeatureFlagOverrideFunction.Disabled"    = true
+    "AzureWebJobs.BulkImportFunction.Disabled"                     = var.disable_bulk_import_function
```
sambiramairelogic, line 164: I know this isn't your change but checking this has made me realise that this line is missing in servicebus_functions and time_functions too. They disable the func in the main resource but not the preview. I know it doesn't hugely matter with it only being the preview slot but would you mind mixing it please? 

(I see this was also true of high_load but that doesn't matter since we're removing that line anyway. I think we need a better system for keeping track of all this because it keeps catching us out...)

ste-banks: Another good spot! Have added them in. We should make it be a part of that PR checklist

sambiramairelogic (approved):

PR #677, opened 2025-05-13
APPT-802: Synchronise global roles and add user manager to admin permissions

sambiramairelogic:
This PR adds `users:manage` to the list of permissions given to admin users. 

It also clears up some inconsistencies between global roles in each environment and adds tests to assert their equivalence. The list of changes is described [on the ticket](https://nhsd-jira.digital.nhs.uk/browse/APPT-802) and also here:

- The system:auditer role has been deleted entirely
- The system:admin-user has been removed from system:all-permissions
- The system-run-reminders and system:run-provisional-sweep roles have been added to system:all-permissions in all environments
- The system:data-importer has been removed from system:admin-user in all environments

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,6 +25,8 @@ stages:
           csvDataToolTestProjectPath: "tests/CsvDataTool.UnitTests/CsvDataTool.UnitTests.csproj"
           bookingExtractsTestProjectPath: "tests/BookingDataExtracts.UnitTests/BookingDataExtracts.UnitTests.csproj"
           capacityExtractsTestProjectPath: "tests/CapacityDataExtracts.UnitTests/CapacityDataExtracts.UnitTests.csproj"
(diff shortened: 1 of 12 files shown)


jsed-nhs:
How does this impact current test/prod users setup? will their be old data in the cosmosDb in each environment? do things need cleaning up?

sambiramairelogic:
> How does this impact current test/prod users setup? will their be old data in the cosmosDb in each environment? do things need cleaning up?

I believe this should be safe, for the following reasons:

- User profiles only store the roles that user has, not the permissions. Permissions are derived from the role.
- I haven't removed any existing roles (with the exception of auditer, which no production user holds).
- I am confident there are no auditers because it's a system role so only devs could've assigned it to anyone
- For roles which have had their permissions changed, that will automatically take effect the next time they log in as these permissions are derived from the role.

Yes, this absolutely will mean re-uploading the global_roles file to each environment. I'm not entirely sure what our process is for doing this in prod but it's a step we'll practice during the staging deployment

ste-banks (approved):

jsed-nhs (approved):

PR #679, opened 2025-05-14
Appt 750 - perf env added

pata9:


infrastructure/resources/resourcegroup.tf
```diff
@@ -1,3 +1,3 @@
 data "azurerm_resource_group" "nbs_mya_resource_group" {
-  name = "${var.application}-rg-${var.environment}-${var.loc}"
+  name = "${var.application}${local.is_perf ? var.application : ""}-rg-${local.is_perf ? "stag" : var.environment}-${var.loc}"
```
pata9, line 2: change to env

data/CosmosDbSeeder/items/perf/core_data/global_roles.json
sambiramairelogic, line 1: Could we add this file to the checks performed in [GlobalRolesTests ](https://github.com/NHSDigital/nbs-appointments-management-service/blob/main/tests/CosmosDbSeederTests/GlobalRolesTests.cs) please?

sambiramairelogic:
We probably need to extend the env parameters for `azure-pipelines-bulk-data` - `azure-pipelines-data-extract` to accept perf too :)

sambiramairelogic:
> We probably need to extend the env parameters for `azure-pipelines-bulk-data` - `azure-pipelines-data-extract` to accept perf too :)

oh and DR

pata9:
> > We probably need to extend the env parameters for `azure-pipelines-bulk-data` - `azure-pipelines-data-extract` to accept perf too :)
> 
> oh and DR

Ye this is a conversation that is needed - 

Extract has never been ran on staging due to the DPS guys not having a mapping environment, might be worth moving the int config to staging? Wouldn't see that as something part of this ticket though, I can get a ticket created

DR - Do we want this on staging or perf or both? Not sure, I was gonna wait on PL for this discussion

data/CosmosDbSeeder/items/perf/core_data/global_roles.json
pata9: Done, updated with merge with your latest stuff :)

PR #682, opened 2025-05-15
APPT-885: Fix Okta provisioning bug

sambiramairelogic:
Attempting to add or edit an Okta user with 24 hours of first creating them is causing an error. 

This is because of a bug in some OktaService logic which assesses if they're provisioned (i.e. invited to okta but not yet activated), and re-activates that provisioning request if it's expired (i.e. they didn't activate their account within 24 hours). This logic was returning an unsuccessful result if they were provisioned but within that 24 hour period, but it should've been returning a success. 

This PR fixes this bug and also re-writes the service to be a bit clearer/more readable.

### DIFF
--- a/src/api/Nhs.Appointments.Core/Okta/OktaService.cs
+++ b/src/api/Nhs.Appointments.Core/Okta/OktaService.cs
@@ -2,32 +2,72 @@ namespace Nhs.Appointments.Core.Okta;
 
 public class OktaService(IOktaUserDirectory oktaUserDirectory) : IOktaService
 {
-    public async Task<UserProvisioningStatus> CreateIfNotExists(string userEmail, string? firstName, string? lastName)
+    public async Task<UserProvisioningStatus> CreateIfNotExists(string userEmail, string firstName, string lastName)
     {
-        var success = false;
-        var oktaUser = await oktaUserDirectory.GetUserAsync(userEmail);
-
-        if (oktaUser != null)
+        var userState = await GetUserState(userEmail);
+        return userState switch
         {
-            var isOlderThanOneDay = DateTime.UtcNow - oktaUser.Created > TimeSpan.FromDays(1);
-
-            if (oktaUser.IsProvisioned && isOlderThanOneDay)
+            UserState.UserDoesNotExist => await CreateUser(userEmail, firstName, lastName),
+            UserState.UserWasProvisionedButOver24HoursAgo => await ReactivateUser(userEmail),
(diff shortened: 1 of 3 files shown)


src/api/Nhs.Appointments.Core/Okta/OktaService.cs
```diff
@@ -1,33 +1,73 @@
...
+    private async Task<UserState> GetUserState(string userEmail)
+    {
+        var user = await oktaUserDirectory.GetUserAsync(userEmail);
+        if (user is null)
+        {
+            return UserState.UserDoesNotExist;
+        }
+
+        var isOlderThanOneDay = timeProvider.GetUtcNow() - user.Created > TimeSpan.FromDays(1);
+        if (user.IsProvisioned && isOlderThanOneDay)
```
jsed-nhs, line 61: Since both of these if's are user.IsProvisioned, can we have that on the outside, then have the single ternary on the inside. i.e `if(provisioned) then old24hr ? status : otherStatus`

src/api/Nhs.Appointments.Core/Okta/OktaService.cs
```diff
@@ -1,33 +1,73 @@
...
+
+    private async Task<UserState> GetUserState(string userEmail)
+    {
+        var user = await oktaUserDirectory.GetUserAsync(userEmail);
+        if (user is null)
+        {
+            return UserState.UserDoesNotExist;
+        }
+
+        var isOlderThanOneDay = timeProvider.GetUtcNow() - user.Created > TimeSpan.FromDays(1);
```
jsed-nhs, line 60: Do we know if the user.created date is UTC? is this from OKTA and is it confirmed somewhere?

jsed-nhs (approved):
Looks fine, two small comments but nothing major.

ste-banks (approved):

src/api/Nhs.Appointments.Core/Okta/OktaService.cs
pata9: if (provisioned) 
return GetProvisionalStatus(user.CreatedDate)

Something like that would be preferred by myself :D can move the time logic into GetProvisionalStatus

src/api/Nhs.Appointments.Core/Okta/OktaService.cs
jsed-nhs: Sam has confirmed to me on Slack that it is UTC and defined in the OKTA docs

jsed-nhs (approved):

pata9 (approved):

PR #683, opened 2025-05-16
APPT-750 - lower staging throughput as perf now owns the high cost

pata9:
Removes the thoughput settings to staging as Perf has been created and now has this throughput. We'll only needed it to be this high for performance testing and as that's been moved to a new environment we can remove it from staging

### DIFF
--- a/infrastructure/environments/stag/main.tf
+++ b/infrastructure/environments/stag/main.tf
@@ -89,16 +89,4 @@ module "mya_application_stag" {
       failover_priority = 1
       zone_redundant    = false
   }]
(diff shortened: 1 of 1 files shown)


DACA16 (approved):

sambiramairelogic (changes requested):
These settings are also set in `stag-ukw`, used for disaster recovery. If we're removing them from uks we should remove them from ukw too :) 

I can't help but notice we also have these setting in prod though, so I have a further concern that if we remove these from staging that's creating a difference between prod and staging.

PR #696, opened 2025-05-22
APPT-814 - Introducing new regional user role

ste-banks:
Introduction of a new regional user role `system:regional-user` which grants a number of permissions to sites within the region
- Regional users **will not** appear in the manage users list if they only have access to a site via the region.
- Regional users are able to manage users, view availability and manage site details for sites within their region but they cannot create availability (unless they're granted that permission on a site level).
- It is still possible for regional users to be granted site level permissions to a site (e.g. Create availability).
- It is possible (at least for now) for users to be granted the regional user permission to multiple regions - there may be further discussions to restrict this to just one region.
- Small refactor of the PermissionChecker class to prevent multiple fetches of permissions in each method. Now it will fetch user roles & role assignments in parallel then filter on them instead. Confirmed all existing unit tests still pass - which they do.
- Added new unit tests for the new region permissions.
- There are no e2e tests yet as there is a further ticket (APPT-868) which involves front end changes for regional users so I figured it was best to leave those for that ticket so we don't have to re-do them.
- Same applies for integration tests on the GetSitesPreviewFunction endpoint - this will be changing as part of APPT-868 as well.

### DIFF
--- a/src/api/Nhs.Appointments.Core/SiteService.cs
+++ b/src/api/Nhs.Appointments.Core/SiteService.cs
@@ -68,12 +68,7 @@ public class SiteService(ISiteStore siteStore, IMemoryCache memoryCache, TimePro
     }
 
     public async Task<IEnumerable<Site>> GetSitesInRegion(string region)
-    {
-        var sites = memoryCache.Get(CacheKey) as IEnumerable<Site>;
-        sites ??= await GetAndCacheSites();
-
-        return sites.Where(s => s.Region == region);
-    }
+        => await siteStore.GetSitesInRegionAsync(region);
 
     public async Task<IEnumerable<SitePreview>> GetSitesPreview()
     {
--- a/src/api/Nhs.Appointments.Core/ISiteStore.cs
+++ b/src/api/Nhs.Appointments.Core/ISiteStore.cs
@@ -17,4 +17,6 @@ public interface ISiteStore
 
     Task<OperationResult> SaveSiteAsync(string siteId, string odsCode, string name, string address, string phoneNumber,
         string icb, string region, Location location, IEnumerable<Accessibility> accessibilities, string type);
+
+    Task<IEnumerable<Site>> GetSitesInRegionAsync(string region);
 }
--- a/src/api/Nhs.Appointments.Persistance/SiteStore.cs
+++ b/src/api/Nhs.Appointments.Persistance/SiteStore.cs
@@ -145,4 +145,9 @@ public class SiteStore(ITypedDocumentCosmosStore<SiteDocument> cosmosStore) : IS
             return new OperationResult(true);
         }
     }
+
+    public Task<IEnumerable<Site>> GetSitesInRegionAsync(string region)
+    {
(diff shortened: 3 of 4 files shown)


src/api/Nhs.Appointments.Core/SiteService.cs
```diff
@@ -67,14 +67,18 @@ public async Task<Site> GetSiteByIdAsync(string siteId, string scope = "*")
         return site;
     }
 
+    public async Task<IEnumerable<Site>> GetSitesInRegion(string region)
+    {
+        var sites = memoryCache.Get(CacheKey) as IEnumerable<Site>;
+        sites ??= await GetAndCacheSites();
```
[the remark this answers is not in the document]
ste-banks: I did that initially, but then saw the `FindSitesByArea` method does this as well so I kept it the same. Happy to add the store method in though.

data/CosmosDbSeeder/items/dev/core_data/global_roles.json
```diff
@@ -114,6 +114,22 @@
...
+        "availability:query",
+        "booking:query",
+        "site:view",
+        "site:view:preview",
+        "site:manage",
+        "site:manage:admin",
+        "site:get-meta-data",
+        "users:view",
+        "users:manage"
+      ]
```
sambiramairelogic, line 132: The permissions in this role match those of `system:admin-user` exactly I believe?  

I think we're blurring concepts here slightly; I see the Role as being "the collection of permissions" and the Scope being "what entities you have those permissions over". By creating a `system:regional-user`, we're blurring those two into one and making the role itself regional. But there's actually nothing to stop us assigning someone the "regional-user" role at a site scope 

I'd personally suggest we re-use the admin role but just assign it with a regional scope

ste-banks: Originally this role didn't have the `booking:query` permission but it was needed to see the _View Availability page_ as it uses the bookings to build up the count. Are we planning to move all of those calculations to the backend eventually with the single source of truth work? If so, this role will need to be updated once again to remove the `booking:query` permission so they can't see the daily appointments page - which would mean they'd be different? If that's not the plan, happy to update.

ste-banks: Further to that, I forgot that we may be adding a new permission around the daily appointments view as well which the admin user will have but a regional one will not. FYI: 
image

sambiramairelogic: Fair enough, I see what you mean. Cos it's called "regional user" I kind of thought we were implying that the role itself was regional, but I guess it's a generic role that just so happens to always be granted at "regional" scope. 

I'd suggest renaming it to avoid this confusion, but I can't think of anything that better describes "the collection of permissions that regional users should have", so... 🤷 

Happy to drop it :)

sambiramairelogic (dismissed):

sambiramairelogic (approved):

PR #704, opened 2025-05-28
APPT-855: Terraform tech debt

sambiramairelogic:
We keep finding more and more discrepancies between which functions are enabled in which environment, or between the main and preview slots. 

This PR moves the onus of declaring this away from the resource files themselves and into the `main.tf` file for each environment. 

Benefits include: 

- Currently if we create a new Function but forget to add it to terraform, it will be enabled in all environments. This PR makes functions disabled by default, so we have to remember to add it to use it. 
- Every main and preview slot has the same, alphabetised list of functions. Because all functions are given an explicit true/false value, the list will be of the same length and order in every resource. This makes it easier to track down future discrepancies. 
- We no longer need the `disable_query_availability_function` or `disable_bulk_import_function` variables
- We no longer need the `create_high_load_function_app` variable. This value is derived by whether the list of high load functions is not empty.


sambiramairelogic:
Still PR has gone stale and if we rediscover the appetite for this work it'll need updating, so closing the PR for now to reduce noise.

PR #706, opened 2025-05-29
APPT-905 - Adding site type to bulk site import - non mandatory field

ste-banks:
- Adding the site type to the site document to be used by the site bulk import - though this field is not mandatory so it will be `null` if the value is not set.
- Haven't added integration tests for this change because we have another ticket (APPT-938) which will cover all bulk import integration tests


src/api/Nhs.Appointments.Core/SiteService.cs
```diff
@@ -80,7 +80,7 @@ public async Task<IEnumerable<SitePreview>> GetSitesPreview()
     }
 
     public async Task<OperationResult> SaveSiteAsync(string siteId, string odsCode, string name, string address, string phoneNumber, string icb,
-        string region, Location location, IEnumerable<Accessibility> accessibilities)
+        string region, Location location, IEnumerable<Accessibility> accessibilities, string? type)
```
sambiramairelogic, line 83: Same here

sambiramairelogic (approved):

PR #707, opened 2025-05-29
APPT-863: Validate coordinates in frontend. APPT-864: Allow empty phone number

sambiramairelogic:
We're not currently validating the Lat/Long co-ordinates when they're edited through MYA itself. 

This is part 1 of 2, which just adds our validation to the frontend. Part 2 will be to add the same validation to the backend API.

Also doubles up as a PR for APPT-864, which is to permit the empty string as a valid entry for the Phone Number field


src/api/Nhs.Appointments.Api/Validators/SetSiteDetailsValidator.cs
sambiramairelogic, line 1: The unit tests for this validator are missing. I've added them to my other PR for this ticket, which is why I'm not testing this change here. 

https://github.com/NHSDigital/nbs-appointments-management-service/pull/709

jsed-nhs (approved):

ste-banks (approved):

PR #710, opened 2025-05-30
APPT-938 - Adding bulk import integration tests

ste-banks:
- Adding user & site bulk import integration tests using the feature toggle pattern introduced by Jonny (hence the target branch).
- The tests cover the on & off states for the BulkImport endpoint for both users and sites.
- These tests don't cover all the validation rules for the user & site endpoints - just making sure these pass for both the on / off states with valid csv files so we have confidence in enabling it
- Added a new base feature step which creates a site with a valid Guid as the BulkImport will reject the file if it is not - have added a TODO to tidy this up later. I was reluctant to make changes to the existing step as it has a lot of uses

### DIFF
--- a/tests/Nhs.Appointments.Api.Integration/Scenarios/BulkImport/SiteBulkImportFeatureSteps.cs
+++ b/tests/Nhs.Appointments.Api.Integration/Scenarios/BulkImport/SiteBulkImportFeatureSteps.cs
@@ -1,0 +1,17 @@
+using Nhs.Appointments.Core.Features;
+using Xunit;
+using Xunit.Gherkin.Quick;
+
+namespace Nhs.Appointments.Api.Integration.Scenarios.BulkImport;
+
+[FeatureFile("./Scenarios/BulkImport/SiteBulkImport_Enabled.feature")]
+public abstract class SiteBulkImportEnabledFeatureSteps(string flag, bool enabled) : BaseBulkImportFeatureSteps(flag, enabled);
+
+[Collection("BulkImportSeriaToggle")]
+public class SiteBulkImport_BulkImportEnabled() : SiteBulkImportEnabledFeatureSteps(Flags.BulkImport, true);
+
+[Collection("BulkImportSeriaToggle")]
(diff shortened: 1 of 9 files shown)


tests/Nhs.Appointments.Api.Integration/Scenarios/BulkImport/SiteBulkImportFeatureSteps.cs
```diff
@@ -0,0 +1,17 @@
+using Nhs.Appointments.Core.Features;
+using Xunit;
+using Xunit.Gherkin.Quick;
+
+namespace Nhs.Appointments.Api.Integration.Scenarios.BulkImport;
+
+[FeatureFile("./Scenarios/BulkImport/SiteBulkImport_Enabled.feature")]
```
jsed-nhs, line 7: I think you can consolidate this down to 2 classes. Each with the FeatureFile and the Collection attrb. You only need to split them further out if this can't be done and other things do/dont need sharing. The two abstract classes can be absorbed, since you don't have an Enabled+Disabled test running over the same file.

tests/Nhs.Appointments.Api.Integration/Scenarios/BulkImport/SiteBulkImportFeatureSteps.cs
```diff
@@ -0,0 +1,17 @@
...
+public abstract class SiteBulkImportEnabledFeatureSteps(string flag, bool enabled) : BaseBulkImportFeatureSteps(flag, enabled);
+
+[Collection("BulkImportSeriaToggle")]
+public class SiteBulkImport_BulkImportEnabled() : SiteBulkImportEnabledFeatureSteps(Flags.BulkImport, true);
+
+[Collection("BulkImportSeriaToggle")]
```
jsed-nhs, line 13: Misnamed the collection value

jsed-nhs (changes requested):

sambiramairelogic (dismissed):

PR #712, opened 2025-06-02
APPT-946 - Removing the check on site ODS code for the site bulk import

ste-banks:


sambiramairelogic:
I think this PR does exactly what the ticket is currently asking for so this isn't a review of your changes per se, but I'm gonna leave a comment on the ticket itself. 
I don't think we should ever be removing all the validation on a bulk upload field, or we'll open ourselves up to all kinds of errors in the future. If we remove validation entirely someone could accidentally put the site name or something in as an ODS code and we wouldn't know until it's live data. 

It looks like ODS codes can only be up to 5 digits long, and comprise only capital letters or numbers. I think we should be validating as such, but will see what product say

sambiramairelogic (changes requested):
Sorry to be a pain, I know I'm possibly being over cautious here but I got agreement from Lauren that we should replace this validation with at least a check that the string comprises only capital letters/number, and doesn't exceed 10 characters. Comment on [ticket](https://nhsd-jira.digital.nhs.uk/browse/APPT-946)

sambiramairelogic (approved):

PR #721, opened 2025-06-06
APPT-920 GetWeekSummary backend function to enable frontend metrics

jsed-nhs:
This PR introduces a new GetWeekSummary backend function to facilitate the source of truth allocation for the front end Week Summary view, so that it accurately reflects MYA's known state.

**Backend Changes**
- New GetWeekSummary function: Returns day-level booking metrics and allocations for a given week start date, to replace the frontend availabilityCalculator.ts logic eventually.
- JSON Response Writer Update: Modified CamelCaseNamingStrategy to avoid processing dictionary keys, preserving expected response formats.
- Session Fetching: Added internal session ID generation when fetching sessions—used to trace allocated slots back to their parent session.
- Allocation Logic Centralisation: BookingAvailabilityStateService now returns WeekSummary using source of truth allocation logic.
- GetOrderedBookings Update: Now accepts booking statuses as parameters to support differing requirements across allocation methods.

**Frontend Changes**
- Feature Flag Support: New endpoint integrated behind the MultipleServices feature flag.
- Return Type Mapping: Mapped the new WeekSummary return type to the existing UI shape to minimise UI refactor.
- Day Card Tests: Jest tests updated to validate functionality with the feature flag toggled.

**Test Improvements**
- Renamed Unit Tests: Clarified naming in BookingAvailabilityState tests for better intent communication.
- New GetWeekSummaryTests: Added dedicated tests to verify greedy allocation logic in weekly summary generation.


src/api/Nhs.Appointments.Api/Functions/GetWeekSummaryFunction.cs
```diff
@@ -0,0 +1,77 @@
...
+        Description = "The id of them site where to get daily availability from")]
+    [OpenApiParameter("from", In = ParameterLocation.Query, Required = true, Type = typeof(double),
+        Description = "The start of the week date")]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.OK, "application/json", typeof(WeekSummary),
+        Description = "A weekly summary for the availability and daily sessions")]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Unauthorized, "application/json",
+        typeof(ErrorMessageResponseItem), Description = "Unauthorized request to a protected API")]
+    [OpenApiResponseWithBody(statusCode: HttpStatusCode.Forbidden, "application/json", typeof(ErrorMessageResponseItem),
+        Description = "Request failed due to insufficient permissions")]
+    [RequiresPermission(Permissions.QueryAvailability, typeof(SiteFromQueryStringInspector))]
```
sambiramairelogic, line 42: Should this not be `Permissions.QueryBooking`? Or a new attribute which can accept multiple permissions and checks for both? 

I'm just thinking that of the two bookings are the more sensitive bit of info you're getting here

src/api/Nhs.Appointments.Api/Functions/GetWeekSummaryFunction.cs
```diff
@@ -0,0 +1,77 @@
...
+    IUserContextProvider userContextProvider,
+    ILogger<GetWeekSummaryFunction> logger,
+    IMetricsRecorder metricsRecorder)
+    : BaseApiFunction<GetWeekSummaryRequest, WeekSummary>(validator, userContextProvider,
+        logger, metricsRecorder)
+{
+    [OpenApiOperation(operationId: "GetWeekSummary", tags: ["Availability"],
+        Summary = "Get weekly availability summary for a week start date and site")]
+    [OpenApiParameter("site", In = ParameterLocation.Query, Required = true, Type = typeof(double),
+        Description = "The id of them site where to get daily availability from")]
```
sambiramairelogic, line 33: Very minor typo in "them" :)

Also, apologies for being a massive pedant but since I'm commenting on this line anyway could we clean up the dangling participle? Perhaps "The ID of the site from which to query bookings and availability"? ;)

src/api/Nhs.Appointments.Api/Validators/GetWeekSummaryRequestValidator.cs
sambiramairelogic, line 1: I've left a general comment on this PR about date ranges, but if we proceed with the Week concept shouldn't this be validating that the date they've passed is a Monday?

src/api/Nhs.Appointments.Api/Functions/GetWeekSummaryFunction.cs
jsed-nhs: I assumed that QueryAvailability would cover permission needed to get both bookings and sessions, as 'Availability' can mean the remaining availability? The NBS QueryAvailability endpoint uses this permission also. Maybe we need a new permission instead for when we want sessions only?

sambiramairelogic: I was thinking of them slightly differently. I've always thought: 

- QueryAvailability is just "can I see the availability documents in the database" 
- QueryBookings is just "can I see the booking documents in the database"

But with there being an implicit suggestion that QueryBookings sort of gives you a bit of QueryAvailability too, since you know that if you can see a Booked booking there must be availability at that time.

src/api/Nhs.Appointments.Core/ScheduleTemplate.cs
```diff
@@ -29,13 +29,26 @@ public class Session
 
 public class SessionInstance : TimePeriod
 {
+    /// <summary>
+    /// Not guaranteed to be generated, only used under certain paths
+    /// </summary>
+    internal Guid? InternalSessionId { get; set; }
```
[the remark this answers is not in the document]
jsed-nhs: Have implemented the second option, with a new class called LinkedSessionInstance. Can discuss more later.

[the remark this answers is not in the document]
jsed-nhs: Initial thoughts are Im not sure this fully solves the issue. The other usages of SessionInstance seem to only be in the SingleService pipelines, and these will eventually be deleted. So the 'improved' new class to not poison the well is likely to be the only class once the dust settles. Worth thinking about more...

src/api/Nhs.Appointments.Api/Functions/GetWeekSummaryFunction.cs
jsed-nhs: Deferring to a future PR for now.

jsed-nhs:
Add in missing function disabled stuff in .tf

src/client/src/app/lib/services/timeService.ts
```diff
@@ -343,3 +343,14 @@ export const addHoursAndMinutesToUkDatetime = (
...
+export const emptyWeekSummary: WeekSummary = {
+  daySummaries: [],
+  maximumCapacity: 0,
+  bookedAppointments: 0,
+  orphanedAppointments: 0,
+  cancelledAppointments: 0,
+  remainingCapacity: 0,
+  startDate: dayjs.utc(),
+  endDate: dayjs.utc(),
+};
```
sambiramairelogic, line 356: It feels a bit odd this living in the `timeService`, I'd have expected it in `@types`. Is it only here because it's using the `dayjs` object? If so, could it accept these dates as ISO strings or something instead?

Furthermore, is there a reason you're using utc here? I thought we outlawed uses of utc in favour of UK time?

src/api/Nhs.Appointments.Core/ScheduleTemplate.cs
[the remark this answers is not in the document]
jsed-nhs: Leaving as is. Tried to do the dictionary generation approach and it isn't any cleaner, and has the added downside of being harder to test.

infrastructure/resources/high_load_functions.tf
```diff
@@ -90,6 +90,7 @@ resource "azurerm_windows_function_app" "nbs_mya_high_load_func_app" {
     "AzureWebJobs.GetPermissionsForUserFunction.Disabled"                  = true
     "AzureWebJobs.GetUserProfileFunction.Disabled"                         = true
     "AzureWebJobs.GetUserRoleAssignmentsFunction.Disabled"                 = true
+    "AzureWebJobs.GetWeekSummaryFunction.Disabled"                         = true
```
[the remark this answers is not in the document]
jsed-nhs: FFS! 👍

src/api/Nhs.Appointments.Core/BookingAvailabilityStateService.cs
```diff
@@ -61,22 +131,82 @@ private async Task<BookingAvailabilityState> BuildState(string site, DateTime fr
...
+
+            //includes both supported and orphaned bookings
+            daySummary.BookedAppointments = bookingsOnDay.Count(x => x.Status == AppointmentStatus.Booked);
+
+            //...of which X are orphaned
+            daySummary.OrphanedAppointments = bookingsOnDay.Count(x =>
+                x.Status == AppointmentStatus.Booked &&
+                x.AvailabilityStatus == AvailabilityStatus.Orphaned);
+
+            daySummary.CancelledAppointments = bookingsOnDay.Count(x => x.Status == AppointmentStatus.Cancelled);
```
sambiramairelogic, line 169: Similarly super pedantic and minor, but unless the compiler rejuggles this we're iterating over `bookingsOnDay` 4 times. Might not read as clearly but doing this in 1 loop should be faster. 

... by like a fraction of a ms, so no real sweat. I think in JS I was doing this was an accumulator, sure dotnet must have an equivalent?

sambiramairelogic (approved):

PR #747, opened 2025-06-17
APPT-992 - Disabling notifications for the user bulk import

ste-banks:
The reason I've added this as an optional parameter instead of splitting out the notifications code is because there was a query around how easy it would be to turn notifications back on for bulk import in the future so it's likely this is only a temporary change.

### DIFF
--- a/tests/Nhs.Appointments.Core.UnitTests/UserServiceTests.cs
+++ b/tests/Nhs.Appointments.Core.UnitTests/UserServiceTests.cs
@@ -393,5 +393,24 @@ namespace Nhs.Appointments.Core.UnitTests
             identityStatus.ExtantInIdentityProvider.Should().BeTrue();
             identityStatus.MeetsWhitelistRequirements.Should().BeTrue();
         }
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


src/api/Nhs.Appointments.Core/UserService.cs
sambiramairelogic, line 1: We're testing that the Function calls this method with sendNotifications = true and that the bulk import calls it with sendNotifications = false, but we're not actually testing that the value of sendNotifications changes the behaviour. Could we add a couple of tests to `UserServiceTests` to cover this? I guess since NotifyRoleAssignmentChanged is a private method, the test could just check whether `bus.Send` is invoked or not?

ste-banks: Sure thing. We've already got a number of tests in there to verify it is called, I'll add one in to make sure it isn't called if SendNotifications is false.

sambiramairelogic: Thanks!

sambiramairelogic (approved):

PR #748, opened 2025-06-17
APPT-956 & APPT-682 RFC Approval step for Production

jsed-nhs:
**This change is a 2-birds-1-stone change - adding the RFC check in the deploy step at the top will prevent the cases where we are having a production lock state due to a timed out terraform plan step on deployment to staging.**

Add EnterChangeRequest validation step in before any of the deploy actions when the environment is Production.

This stops the terraform plan step from running (and acquiring a Production state lock) when a release is deployed to Staging (since it won't kick off the plan step without this manual validation passing). Once this validation has timed out, the Production stage can be reran (on the day of LIVE deployment) and if the validation passes with an entered CFG, then it will do the plan and acquire the lock and hopefully be successfully deployed and then release the lock.

This is a temporary workaround while we try and fine-tune our process. This comment field can't be validated for valid Regex (or anything for that matter!) and a workaround would have to be added for it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,7 +7,7 @@ pool:
 parameters:
   - name: environments
     type: object
-    default: ["dev", "int"]
+    default: ["dev", "int", "stag", "prod"]
 
 variables:
   prBuildServiceConnectionName: "nbs-myacicd-rg-int"
@@ -339,5 +339,5 @@ stages:
           cosmosAccountName: nbs-mya-cdb-${{env}}-uks
           resourceGroup: nbs-mya-rg-${{env}}-uks
           serviceConnectionName: nbs-mya-rg-${{env}}
-          isMain: true
+          isMain: ${{eq(variables['Build.SourceBranch'], 'refs/heads/main')}}
           buildNumber: "$(Build.BuildNumber)"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,7 +7,7 @@ pool:
 parameters:
   - name: environments
     type: object
-    default: ["dev", "int", "stag", "prod"]
+    default: ["dev", "int"]
 
 variables:
   prBuildServiceConnectionName: "nbs-myacicd-rg-int"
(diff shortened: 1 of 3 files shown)


sambiramairelogic:
Do we not also observe the state lock errors when deploying to staging? I thought this was an issue on all environments?

I also still think we'd run into the issue if we took over 1hr to approve the terraform plan

sambiramairelogic:
> > Do we not also observe the state lock errors when deploying to staging? I thought this was an issue on all environments?
> > I also still think we'd run into the issue if we took over 1hr to approve the terraform plan
> 
> We do, but we can manually release those ourselves since we have permission to do so if needed. Yes we would, however with this change the plan should only be run once on the day of deployment and should be reviewed and passed as urgency; we shouldn't be taking over an hour to review a plan on a LIVE deployment.

I think we could easily take over an hour on the plan. It might not be for technical reasons; we might have process-related delays or delays because someone "just wants to check something first...". Just judging by how long it took us the other week. If we've changed infrastructure we could easily take 10-15 minutes on that as devs, and 45 mins could easily get lost to process in my opinion. 

I'm not objecting to the above btw, just saying I don't think it fixes the lock issue

jsed-nhs:
> > > Do we not also observe the state lock errors when deploying to staging? I thought this was an issue on all environments?
> > > I also still think we'd run into the issue if we took over 1hr to approve the terraform plan
> > 
> > 
> > We do, but we can manually release those ourselves since we have permission to do so if needed. Yes we would, however with this change the plan should only be run once on the day of deployment and should be reviewed and passed as urgency; we shouldn't be taking over an hour to review a plan on a LIVE deployment.
> 
> I think we could easily take over an hour on the plan. It might not be for technical reasons; we might have process-related delays or delays because someone "just wants to check something first...". Just judging by how long it took us the other week. If we've changed infrastructure we could easily take 10-15 minutes on that as devs, and 45 mins could easily get lost to process in my opinion.
> 
> I'm not objecting to the above btw, just saying I don't think it fixes the lock issue

I agree, I don't think this is an 'this issue is resolved once and for all' fix, I just think it gets us in a better place for the next release where we aren't 100% dependent on requesting access to release the lock again.

PR #749, opened 2025-06-18
APPT-990 - Conditionally register okta user role change consumer

Dave-O-NHS:


sambiramairelogic (dismissed):

src/api/Nhs.Appointments.Api/Functions/NotifyOktaUserRolesChangedFunction.cs
```diff
@@ -5,16 +5,25 @@
...
+        [ServiceBusTrigger(QueueName, Connection = "ServiceBusConnectionString")] 
+        ServiceBusReceivedMessage message, 
+        CancellationToken cancellationToken
+    ){
+        var isOktaEnabled = await featureToggleHelper.IsFeatureEnabled(Flags.OktaEnabled);
+
+        if (isOktaEnabled)
+        {
+            await receiver.HandleConsumer<OktaUserRolesChangedConsumer>(QueueName, message, cancellationToken);
+        }
```
sambiramairelogic, line 27: A pattern we've started following elsewhere is to return a 501 NotImplemented when something is toggled off, so could optionally return that here. An added bonus would be that the event gets dead-lettered so it could be re-ran if we later turned the toggle on

Dave-O-NHS: I will look into that 501 response, thanks

sambiramairelogic (approved):

PR #758, opened 2025-06-23
APPT-XXX Update E2E tests to avoid date collisions

jsed-nhs:
We have hard-coded local seed data that is colliding with current increment test, kick the can down the road...


sambiramairelogic (approved):
Approved, but just a side-thought about the 340 day increment is obviously we can't create availability more than a year in advance. So if someone added a test expecting to see an Add Session button for every day in that month or something we could run into issues, but I presume these tests don't look at month view so are passing on 340.

PR #764, opened 2025-06-25
APPT-868 - Update site list page for regional users

Dave-O-NHS:
# Description

Re-design of sites list page includes:
- new search bar with search and clear buttons
- table displaying sites was extended with ICB data

Fixes # (issue)
image

# Checklist:


ste-banks (changes requested):
I would recommend we rebase this branch off mine `appt-869/site-list-additions` and re-target this PR to that branch because there is duplicated code with that PR.

Update: I'd probably revert all e2e test changes in this PR around the new table now this is rebased off my branch as they all pass and were all approved and this PR is failing a build.

src/client/src/app/lib/components/site-list.tsx
```diff
@@ -13,40 +12,60 @@ type Props = {
...
-        aria-label="site-search"
-        placeholder="Search"
-        onChange={debounceSearchHandler}
-      ></TextInput>
+      <div className="search-bar">
+        <div>
+          <TextInput
+            id="site-search"
+            label="Search active sites by name, ICB or ODS code"
+            aria-label="site-search"
```
sambiramairelogic, line 45: Aria-labels will be read out by screen readers so they should be human readable, similar to visual labels. 

The Content team should really be specifying what our aria-labels should be so you could ask them, or just make it the same as the label. I don't mind which :)

sambiramairelogic (commented):
It would be good to get some Playwright tests in for the site search, as it's a user journey all by itself really

Dave-O-NHS:
> It would be good to get some Playwright tests in for the ODS / ICB fields in site search, as it's a user journey all by itself really

sambiramairelogic (approved):

ste-banks (approved):

PR #767, opened 2025-06-26
APPT-1021 Orphaned total is double counted bug fix

jsed-nhs:
# Description

In the front end, the booked metric is both the supported bookings and the orphaned. I reflected this in my API metric. But the UI adds these two together, so its easier for me to only count supported bookings in the API for consistency and keep allowing the front end to add the two totals together. 

Fixes # (issue)

# Checklist:

### DIFF
--- a/data/CosmosDbSeeder/items/local/booking_data/seed_booking_35.json
+++ b/data/CosmosDbSeeder/items/local/booking_data/seed_booking_35.json
@@ -3,7 +3,7 @@
   "from": "2025-08-20T11:25:00",
   "duration": 5,
   "service": "COVID_FLU:18_64",
-  "created": "2025-10-21T06:23:46.654321+00:00",
+  "created": "2025-07-21T06:23:46.654321+00:00",
   "status": "Booked",
   "availabilityStatus": "Orphaned",
   "statusUpdated": "0001-01-01T00:00:00+00:00",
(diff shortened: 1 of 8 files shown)


data/CosmosDbSeeder/items/local/booking_data/seed_booking_35.json
```diff
@@ -0,0 +1,28 @@
+{
+  "reference": "seed-booking-reference-35",
+  "from": "2025-08-20T11:25:00",
+  "duration": 5,
+  "service": "COVID_FLU:18_64",
```
sambiramairelogic, line 5: I guess actually, now I come to think of it, we're not addressing this data in our feature toggles tests. For Playwright tests where multiple services is OFF, we technically shouldn't have this data in the db. God knows how we'd tackle that though... Well, I guess we'd need to do what I've been suggesting for awhile and have the Playwright's setup/teardown routines run the seeder with clear-container true

jsed-nhs: Agreed, this currently isn't an issue though as the playright test thta uses this data only runs for MS enabled.

sambiramairelogic (approved):

pata9 (approved):

data/CosmosDbSeeder/items/local/booking_data/seed_booking_35.json
```diff
@@ -0,0 +1,28 @@
...
+  "created": "2025-10-21T06:23:46.654321+00:00",
```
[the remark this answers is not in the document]
pata9: Can definitely look to explore this!

PR #768, opened 2025-06-27
Appt 923 - Daily appointments page permission

ste-banks:
# Description

Added a new permission `booking:view-detail` and applied it to availability manager, appointment manager, site details manager, user manager and admin roles. It **has not** been applied to the regional manager role.

Fixes # (issue)

# Checklist:

### DIFF
--- a/data/CosmosDbSeeder/items/dev/core_data/global_roles.json
+++ b/data/CosmosDbSeeder/items/dev/core_data/global_roles.json
@@ -9,7 +9,7 @@
       "permissions": [
         "availability:setup",
         "availability:query",
-        "booking:list-view",
+        "booking:view-detail",
         "booking:query",
         "site:view",
         "site:view:preview",
@@ -23,7 +23,7 @@
       "permissions": [
         "availability:query",
         "booking:cancel",
-        "booking:list-view",
+        "booking:view-detail",
         "booking:query",
         "site:view",
         "site:view:preview",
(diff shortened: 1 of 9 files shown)


jsed-nhs (changes requested):
Looks fine to me, just wondering if we rename the permission to something like "booking:view:detail"? I think 'list' is too vague to me, what the permission sounds like its controlling is viewing booking details (as we are restricting users viewing them at that detail). I think list is vague because you could technically argue you are seeing booking list metrics on the other views? Not sure if I'm being too pedantic :D

jsed-nhs:
> > Looks fine to me, just wondering if we rename the permission to something like "booking:view:detail"? I think 'list' is too vague to me, what the permission sounds like its controlling is viewing booking details (as we are restricting users viewing them at that detail). I think list is vague because you could technically argue you are seeing booking list metrics on the other views? Not sure if I'm being too pedantic :D
> 
> The naming `booking:list-view` was defined on the ticket. Though I do think it's a reasonable name because that's exactly what the page is in essence, a list of bookings :D

Sorry - not sure I've made myself clear. I think tying a permission to a type of page doesn't feel right to me, it should be about the permission to data and the level to which you have access to it. If we ever changed it from a list to a different style of page UI, the permission name is then incorrect. If we ever have a page to a specific booking with details outside of a list, you may want to control that via this permission also. From reading the ticket, it sounds like the access is preventing users viewing booking details, so thats what I feel the permission should reflect. All of our other permissions are around user actions, not site pages. i.e query bookings, cancel booking, manage sites, manage users etc - rather than describing pages the user has access to.

PR #771, opened 2025-07-01
Appt 870 - Site search updates

ste-banks:
# Description

Site search updates to search by site name or ODS code. Most of the styling for this had already mostly been done as part of David O's PR for APPT-868, I've just made a small change to make the buttons match the screenshots in the tickets.

This also includes an update to all secondary buttons in the app to give them the `nhsuk-button--secondary-solid` class after the recent design package update.

Fixes # (issue)

# Checklist:


jsed-nhs (changes requested):
'AdditionalClasses' seems like an anti-pattern to me. Would rather have a specific 'search-button' component, or at the very least if following this pattern that the class is applied at the top level (not drilled down) and the Button class appends its final class onto the existing classNames.

PR #774, opened 2025-07-02
APPT-1027: Changing time in capacity extract to 24h format

Dave-O-NHS:
# Description

Fixes # Time data in extracts is not in 24h format to await confusion. 

# Checklist:

### DIFF
--- a/data/CapacityDataExtract/Program.cs
+++ b/data/CapacityDataExtract/Program.cs
@@ -14,7 +14,7 @@ builder.Configuration
             .AddNbsAzureKeyVault();
 
 builder.Services
-    .AddDataExtractServices("capacity", builder.Configuration, args.Contains("create-local-sample"))
+    .AddDataExtractServices("BookingCapacity", builder.Configuration, args.Contains("create-local-sample"))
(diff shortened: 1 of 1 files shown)


data/CapacityDataExtract/Program.cs
```diff
@@ -14,7 +14,7 @@
             .AddNbsAzureKeyVault();
 
 builder.Services
-    .AddDataExtractServices("BookingCapacity", builder.Configuration, args.Contains("create-local-sample"))
```
sambiramairelogic, line 17: I think this will cause errors in `CapacityExtracts.feature` because it's still looking for a file with the old name. It'd also be good to check with Tom Batchelor / @pata9 if this will affect the reporting, as I think they're looking for a file with this name too?

pata9: Ye we can't change the name of the file, it's dictated by DPS team. Reason the test exists :)

Dave-O-NHS: OK sorry for that, will update acceptance criteria and revert file name.

PR #775, opened 2025-07-02
APPT-983: Increasing container width to 1100px

Dave-O-NHS:
# Description

Increasing container width to 1100px to allows us to display more content to user.  

# Checklist:


sambiramairelogic (changes requested):
I don't think we should be creating manual CSS classes if we can avoid it. The nhsuk-frontend package has a -wide setting for this class which we should be using instead. The JIRA ticket itself says that an upgrade to this package is included in its scope, so we can nail that bit of tech debt at the same time :)

PR #776, opened 2025-07-02
APPT-1035 Update change session logic to be future Calendar date

jsed-nhs:
# Description

Update session summary table change action link to only appear when the session is in a future calendar date, rather than just in the future. This aligns with the backend API validation and the logic for creating new sessions.

Deleted July 2025 local seed data as is not used and causing e2e test issues.

Fixes # (issue)

# Checklist:


sambiramairelogic (changes requested):
As discussed these dates were deliberately close to force us to address this tech debt issue. It's actually one of the ones I brought up yesterday that we've just created a ticket for. I'm concerned that if buy ourselves a whole extra year, we'll lose any support from planning to address this. Could we only bump these dates along by 1 month please?

sambiramairelogic (approved):

PR #782, opened 2025-07-04
APPT-950 - Removing the cancelled appointment count on week & day view

ste-banks:
# Description

Removing the cancellation calculation on week & day view cards. The manually cancelled calculation **will remain**.

Fixes # (issue)

# Checklist:


jsed-nhs (changes requested):
Suspect my new GetWeekSummaryFunction will need updating for this change. Sounds like it no longer needs to return the cancelled metrics, and also doesn't need to fetch cancelled bookings. Will need the tests updating once this is changed.

jsed-nhs (approved):
Made the required change myself, but need someone else to ok my work

PR #803, opened 2025-07-16
APPT-750: Create performance environment (2)

sambiramairelogic:
# Description

Creates the terraform, pipeline and CosmosDBSeeder changes necessary to create the performance testing environment for MYA. Unlike other environments, this does not live on its own subscription but shares one with Staging. Hence rather than follow the resource group naming convention used elsewhere (`nbs-mya-rg-{env}-uks`), the resource group for Perf is named `nbs-myaperf-rg-stag-uks` to avoid naming conflicts and adhere to the NHS' rules about resource group names.

This builds atop the [original PR](https://github.com/NHSDigital/nbs-appointments-management-service/pull/679) by @pata9 which had 96 commits and became too much trouble to rebase due to conflicts. 99% of the work here is Paul's! :)

Fixes # ([issue](https://nhsd-jira.digital.nhs.uk/browse/APPT-750))

# Checklist:


features/perf.feature.flags.json
```diff
@@ -0,0 +1,8 @@
+{
+  "FeatureManagement": {
```
pata9, line 2: Should probs align this with prod? But maybe double check what PL needs for his current tests, I think he's moving onto 2.3 now

pata9 (approved):

pata9 (approved):

features/perf.feature.flags.json
sambiramairelogic: Yeah good point, I just naively copied across from the other PR. I guess there's a confluence doc somewhere saying what these should be, I'll see if I can find one...

pata9 (dismissed):

features/perf.feature.flags.json
sambiramairelogic: Updated

pata9 (approved):

PR #837, opened 2025-07-28
APPT-1147: Unable to add okta user

Dave-O-NHS:
# Description

We are fixing 2 issues. 
First one was to convert status mapping switch statement to use userStatus enum value instead of name because of inconsistency in okta enum naming.  
Second issue was that in case of deactivated account, we were trying to create account instead of reactivate and that was failing. 

# Checklist:


jsed-nhs (changes requested):
The unit test looks a bit suspicious to me

jsed-nhs (approved):

PR #856, opened 2025-08-04
Reports/performance

pata9:
# Description

Added 'chunking' to the trigger. The trigger will now run multiple times in a day and publish x days at a time.

Fixes # (issue)

# Checklist:


src/api/Nhs.Appointments.Core/Messaging/MassTransitBusWrapper.cs
```diff
@@ -6,9 +6,9 @@ public class MassTransitBusWrapper(IBus bus) : IMessageBus
 {
     public async Task Send<T>(params T[] messages) where T : class
     {
-       foreach(T message in messages)
-        {
-            await bus.Send(message);
-        }
+        EndpointConvention.TryGetDestinationAddress<T>(out var destinationAddress);
```
jsed-nhs, line 9: We doing batching by default for all messaging?

jsed-nhs (changes requested):

jsed-nhs (approved):

PR #865, opened 2025-08-06
APPT-1203 SitesByArea SitesSupportService filtering

jsed-nhs:
# Description

Update to the GetSitesByAreaFunction to allow new optional filtering to be used by NBS to restrict the long/lat area results to only return sites that support the service being queried (this means that a site has posted ANY availability within the daterange, it doesn't mean they have capacity remaining - i.e we aren't doing full allocation and free slot checks at this top level stage)

This is supposed to be 'temporary' in that we are going to attempt to consolidate this API endpoint and the first api/query call into a new single API endpoint in future (APPT-841)

**Having said that, the intention is for this to go out as a hotfix to address A/W needs, so it needs to be performant and behave as expected.**

Have added a LOT of integration tests to prove expected behaviours, as well as old behaviour is still supported if needed to be rolled back to by NBS.

Have added logging in for the 'Batching' of calls so that we can verify in logs in Production whether the batch size needs increasing/decreasing for certain services...?

The API does 2 calls, one to get all the cached sites in API and calculated distance bounds in memory. Then I use all the potentially valid sites (ordered distance ascending) to try and batch build up the required results. If needed, it can iterate multiple times to find a site that is the furthest away (whilst still within range) if its the only one to support the service.

Fixes # (issue)

# Checklist:

### DIFF
--- a/src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
+++ b/src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
@@ -46,11 +46,9 @@ public class GetSitesByAreaFunction(
         Description =
             "Optional parameter to be used alongside from and until. When all three of these parameters are used, it filters the results to only those sites that support the services within that date range")]
     [OpenApiParameter("from", In = ParameterLocation.Query, Required = false, Type = typeof(string),
-        CollectionDelimiter = OpenApiParameterCollectionDelimiterType.Comma,
         Description =
             "Optional parameter to be used alongside services and until. When all three of these parameters are used, it filters the results to only those sites that support the services within that date range. DateFormat yyyy-MM-dd.")]
     [OpenApiParameter("until", In = ParameterLocation.Query, Required = false, Type = typeof(string),
-        CollectionDelimiter = OpenApiParameterCollectionDelimiterType.Comma,
         Description =
             "Optional parameter to be used alongside services and from. When all three of these parameters are used, it filters the results to only those sites that support the services within that date range. DateFormat yyyy-MM-dd.")]
     [OpenApiResponseWithBody(statusCode: HttpStatusCode.OK, "application/json", typeof(IEnumerable<SiteWithDistance>),
--- a/src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
+++ b/src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
@@ -73,16 +73,14 @@ public class GetSitesByAreaFunction(
     {
         SiteSupportsServiceFilter siteSupportsServiceFilter = null;
 
-        //should pass validation rules
-        if (request.services is { Length: 1 } && request.from != null && request.until != null)
+        //if all 3 params are provided correctly, use the SiteSupportsServiceFilter
+        if (request.Services is { Length: 1 } && request.FromDate != null && request.UntilDate != null)
         {
-            var fromDate = DateOnly.ParseExact(request.from, "yyyy-MM-dd");
-            var untilDate = DateOnly.ParseExact(request.until, "yyyy-MM-dd");
-            siteSupportsServiceFilter = new SiteSupportsServiceFilter(request.services.Single(), fromDate, untilDate);
+            siteSupportsServiceFilter = new SiteSupportsServiceFilter(request.Services.Single(), request.FromDate.Value, request.UntilDate.Value);
         }
 
-        var sites = await siteService.FindSitesByArea(request.longitude, request.latitude, request.searchRadius,
-            request.maximumRecords, request.accessNeeds, request.ignoreCache, siteSupportsServiceFilter);
+        var sites = await siteService.FindSitesByArea(request.Longitude, request.Latitude, request.SearchRadius,
+            request.MaximumRecords, request.AccessNeeds, request.IgnoreCache, siteSupportsServiceFilter);
         return ApiResult<IEnumerable<SiteWithDistance>>.Success(sites);
     }
 
--- a/src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
+++ b/src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
@@ -9,13 +9,10 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Attributes;
 using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Enums;
-using Microsoft.Azure.WebJobs.Extensions.OpenApi.Core.Extensions;
 using Microsoft.Extensions.Logging;
 using Microsoft.OpenApi.Models;
-using Nhs.Appointments.Api.Auth;
 using Nhs.Appointments.Api.Models;
 using Nhs.Appointments.Core;
-using Nhs.Appointments.Core.Inspectors;
 
 namespace Nhs.Appointments.Api.Functions;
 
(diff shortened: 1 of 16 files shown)


src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
```diff
@@ -41,6 +41,18 @@ public class GetSitesByAreaFunction(
     [OpenApiParameter("accessNeeds", In = ParameterLocation.Query, Required = false, Type = typeof(string[]),
         CollectionDelimiter = OpenApiParameterCollectionDelimiterType.Comma,
         Description = "Required access needs used to filter sites")]
+    [OpenApiParameter("services", In = ParameterLocation.Query, Required = false, Type = typeof(string[]),
+        CollectionDelimiter = OpenApiParameterCollectionDelimiterType.Comma,
+        Description =
+            "Optional parameter to be used alongside from and until. When all three of these parameters are used, it filters the results to only those sites that support the services within that date range")]
+    [OpenApiParameter("from", In = ParameterLocation.Query, Required = false, Type = typeof(string),
+        CollectionDelimiter = OpenApiParameterCollectionDelimiterType.Comma,
```
sambiramairelogic, line 49: I don't think you need the CollectionDelimiter lines for the from/until dates as they're just strings

src/api/Nhs.Appointments.Core/IAvailabilityStore.cs
```diff
@@ -5,6 +5,9 @@ public interface IAvailabilityStore
     Task<IEnumerable<SessionInstance>> GetSessions(string site, DateOnly from, DateOnly to);
    Task ApplyAvailabilityTemplate(string site, DateOnly date, Session[] sessions, ApplyAvailabilityMode mode,
         Session sessionToEdit = null);
-    Task<IEnumerable<DailyAvailability>> GetDailyAvailability(string site, DateOnly from, DateOnly to);
-    Task<SessionInstance> CancelSession(string site, DateOnly date, Session session);
+   Task<IEnumerable<DailyAvailability>> GetDailyAvailability(string site, DateOnly from, DateOnly to);
+   Task<SessionInstance> CancelSession(string site, DateOnly date, Session session);
+
+   Task<bool> SiteSupportsService(string siteId, string service,
+       List<string> dailyAvailabilityIds);
```
sambiramairelogic, line 12: Could I just suggest some minor naming tweaks here? 

I know since you're passing dates in it's kind of implicit that you're checking a range, but that param is called Id so this method actually looks like a global on/off sort of thing. I think something like `SiteOffersServiceDuringPeriod` would convey the content more clearly.
Likewise, instead of `dailyAvailabilityIds`, I'd call it something more like `DatesInPeriod`. The fact that the Id of our cosmos documents is the date string is a facet of our implementation, and by naming this param Ids we bake that into this method. I'd rather describe them a dates because that's clearer to the caller and agnostic for the future

src/api/Nhs.Appointments.Core/SiteService.cs
```diff
@@ -43,14 +45,91 @@ public async Task<IEnumerable<SiteWithDistance>> FindSitesByArea(double longitud
...
+        var orderedSites = sites.OrderBy(site => site.Distance).ToList();
+        
+        var results = new List<SiteWithDistance>();
+
+        var generatedIds = GetDateStringsInRange(from, to);
+        var iterations = 0;
+        
+        //while we are still short of the max, keep appending results
+        //ideally, the first batch would contain more than or equal to the max results, so won't need to iterate often...
+        while (results.Count < maxRecords)
```
sambiramairelogic, line 83: Almost certainly fine, but just a rogue thought. 
I don't think we're restricting the date range that can be requested. I'm just thinking the worst case scenario here is we iterate through every single site in the system over like a 20 year period. 

Should we:
- Restrict the date range so `from` can't be before today, and `until` can't be more than 6 months in the future?
- Limit the max (iterations * batchsize) so we're not querying more than, say, 500 sites? If MYA had 100k sites in the future we wouldn't wanna query all of them

jsed-nhs: We don't currently validate the date range at the mo, could be worth adding in if agreed. We currently do restrict maxRecords in the call to max 50, but I spose if another caller used this service for something else that may need adressing...?

src/api/Nhs.Appointments.Core/SiteService.cs
```diff
@@ -43,14 +45,91 @@ public async Task<IEnumerable<SiteWithDistance>> FindSitesByArea(double longitud
...
+        {
+            var concurrentBatchResults = new ConcurrentBag<SiteWithDistance>();
+            
+            var orderedSiteBatch = orderedSites.Skip(iterations * batchSize).Take(batchSize).ToList();
```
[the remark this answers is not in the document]
jsed-nhs: I think I might leave this for now to see how it performs, we can always improve it if required.

src/api/Nhs.Appointments.Api/Validators/GetSitesByAreaRequestValidator.cs
```diff
@@ -33,32 +33,32 @@ public GetSitesByAreaRequestValidator()
...
                             .WithMessage("'From' must be a date in the format 'yyyy-MM-dd'")
-                            .Must(x => DateOnly.TryParseExact(x.until, "yyyy-MM-dd", out _))
+                            .Must(x => DateOnly.TryParseExact(x.Until, "yyyy-MM-dd", out _))
                             .WithMessage("'Until' must be a date in the format 'yyyy-MM-dd'")
                             .Must(x =>
                             {
-                                var fromDate = DateOnly.ParseExact(x.from, "yyyy-MM-dd");
-                                var untilDate = DateOnly.ParseExact(x.until, "yyyy-MM-dd");
+                                var fromDate = DateOnly.ParseExact(x.From, "yyyy-MM-dd");
+                                var untilDate = DateOnly.ParseExact(x.Until, "yyyy-MM-dd");
```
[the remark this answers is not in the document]
jsed-nhs: The computed property will throw an exception if not parsed properly.

src/api/Nhs.Appointments.Api/Validators/GetSitesByAreaRequestValidator.cs
```diff
@@ -7,21 +8,61 @@ public class GetSitesByAreaRequestValidator : AbstractValidator<GetSitesByAreaRe
...
+                            .WithMessage("'From' must be a date in the format 'yyyy-MM-dd'")
+                            .Must(x => DateOnly.TryParseExact(x.Until, "yyyy-MM-dd", out _))
+                            .WithMessage("'Until' must be a date in the format 'yyyy-MM-dd'")
+                            .Must(x =>
+                            {
+                                var fromDate = DateOnly.ParseExact(x.From, "yyyy-MM-dd");
+                                var untilDate = DateOnly.ParseExact(x.Until, "yyyy-MM-dd");
+                                return untilDate >= fromDate;
+                            })
+                            .WithMessage("'Until' date must be greater than or equal to 'From' date");
```
sambiramairelogic, line 64: As I've said somewhere else, think you can make this a lot simpler now you don't need to parse the dates here

src/api/Nhs.Appointments.Core/SiteService.cs
ste-banks: On the date range, IIRC you can't book the same day so you could ensure it's tomorrow onwards and doesn't NBS only look 25 days into the future? (though possibly raising to 30 days?) so we could restrict it the same way (but make it configurable) to improve performance.

sambiramairelogic (approved):
Provisionally approved as comments are nonblocking

src/api/Nhs.Appointments.Core/SiteService.cs
jsed-nhs: Leaving alone for now as no requirements to update existing validation rules

src/api/Nhs.Appointments.Api/Functions/GetSitesByAreaFunction.cs
```diff
@@ -52,16 +62,25 @@ public class GetSitesByAreaFunction(
...
     {
-        var sites = await siteService.FindSitesByArea(request.longitude, request.latitude, request.searchRadius,
-            request.maximumRecords, request.accessNeeds, request.ignoreCache);
+        SiteSupportsServiceFilter siteSupportsServiceFilter = null;
+
+        //if all 3 params are provided correctly, use the SiteSupportsServiceFilter
+        if (request.Services is { Length: 1 } && request.FromDate != null && request.UntilDate != null)
+        {
+            siteSupportsServiceFilter = new SiteSupportsServiceFilter(request.Services.Single(), request.FromDate.Value, request.UntilDate.Value);
+        }
```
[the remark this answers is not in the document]
jsed-nhs: Can always refactor in main alone if devs feel strongly enough for neatness in future.

PR #895, opened 2025-08-13
APPT-1249 RSV/FLU2:3 validation and memory caching

jsed-nhs:
# Description

To ensure NBS do not use this potentially non-performant endpoint for busier services, the new optional parameters can only be used when 'RSV:Adult' and 'Flu:2_3' are provided (separately).

Additionally, I have realised a lot of the slowness will have come from the fact that we aren't caching these expensive operations, and this can be acheived by setting a key for the specific site, service and date range (with value of true/false whether the site supports that service). This should improve performance significantly.

Have also upped the batchSize in iterations as, when coupled with the new caching, should find results much faster.

Fixes # (issue)

# Checklist:

### DIFF
--- a/tests/Nhs.Appointments.Api.UnitTests/Validators/GetSitesByAreaRequestValidatorTests.cs
+++ b/tests/Nhs.Appointments.Api.UnitTests/Validators/GetSitesByAreaRequestValidatorTests.cs
@@ -168,9 +168,6 @@ public class GetSitesByAreaRequestValidatorTests
         result.Errors.Single().ErrorMessage.Should().Be("'Services' currently only supports: 'RSV:Adult or 'FLU:2_3'");     
     }
     
-    /// <summary>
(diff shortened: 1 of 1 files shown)


src/api/Nhs.Appointments.Api/Validators/GetSitesByAreaRequestValidator.cs
```diff
@@ -47,8 +48,8 @@ public GetSitesByAreaRequestValidator()
                     () =>
                     {
                         RuleFor(x => x.Services)
-                            .Must(services => services.Length == 1)
-                            .WithMessage("'Services' currently only supports one service");
+                            .Must(services => services.Length == 1 && (services.Single() == "RSV:Adult" || services.Single() == "FLU:2_3"))
```
pata9, line 51: We need to think this one through, okay for hotfix but not long term

pata9 (approved):

PR #903, opened 2025-08-15
APPT-XXX: allow us to deploy main line build to perf and pen if we need to

pata9:
# Description

Starting to be a real bug bear of mine, we should be able to do one build to as many environments as possible. I don't want to have to spin up another build, especially with the standing E2E commit conflicts

Fixes # (issue)

# Checklist:


jsed-nhs (changes requested):
Disagree with auto deploy to perf env on mainline merge

PR #922, opened 2025-08-20
Appt 1184 - Bulk import handle ICB scoped permissions

ste-banks:
# Description

- Additional column `ICB` in the user upload csv file and it's a non-mandatory field
- Only one ICB scoped permission per user is allowed (duplicates will cause the upload to be rejected)
- ICB code must be in the well known ods codes document - otherwise the upload will be rejected
- If a new ICB scoped permission for a user, add it
- If a user has an existing ICB scoped permission and the csv contains an ICB scoped permission to a different ICB, override it
- If a user has an existing ICB scoped permission and the csv contains the same one, remove the permission
- Remove an existing region scoped permission if adding an ICB scoped permission
- Update to the regional permissions - removing an ICB scoped permission if adding a regional scoped permission
- Unit tests for all of the above and an update to the integration test set up for user bulk upload

Fixes # (issue)

# Checklist:


src/api/Nhs.Appointments.Core/BulkImport/UserDataImportHandler.cs
```diff
@@ -46,11 +46,30 @@ public async Task<IEnumerable<ReportItem>> ProcessFile(IFormFile inputFile)
...
+
+            ValidateRegionCodes(regionalUsers, report, validRegionCodes);
+            CheckForDuplicatedPermissions(regionalUsers, report, "region");
+        }
+
+        var icbUsers = userImportRows.Where(usr => !string.IsNullOrEmpty(usr.Icb)).ToList();
+        if (icbUsers.Count > 0)
+        {
+            var validIcbCodes = wellKnownOdsCodes
+                .Where(ods => ods.Type.Equals("icb", StringComparison.CurrentCultureIgnoreCase))
```
pata9, line 67: Food for thought, I'd normally use `OrdinalIgnoreCase` and we do in other places in this file by the looks of things. Don't think it will have much effect here, but here's an article on it https://benscabbia.co.uk/programming/ordinal-vs-culture-comparison-in-csharp

src/api/Nhs.Appointments.Persistance/UserStore.cs
```diff
@@ -163,4 +173,53 @@ public async Task SaveAdminUserAsync(User adminUser)
...
+
+        var existingRoleAssignments = originalDocument.RoleAssignments;
+        var updatedRoleAssignments = existingRoleAssignments.AsEnumerable();
+
+        var hasIcbScopedPermission = updatedRoleAssignments.Any(ra => ra.Scope == scope);
+        var hasOtherIcbScopedPermission = updatedRoleAssignments.Any(ra => ra.Role == IcbUserRole && ra.Scope != scope);
+        var hasRegionalPermission = updatedRoleAssignments.Any(ra => ra.Role == RegionalUserRole);
+
+        // User already has this ICB permission - therefore this is treated as a removal
+        if (hasIcbScopedPermission)
```
pata9, line 199: I think we need a few more test cases around this, a lot of conditions in here, I don't think we can cover them in 2 tests?

src/api/Nhs.Appointments.Persistance/UserStore.cs
```diff
@@ -163,4 +173,53 @@ public async Task SaveAdminUserAsync(User adminUser)
...
+                .Concat(roleAssignments);
+        }
+        // No existing ICB permission - add it
+        else
+        {
+            updatedRoleAssignments = updatedRoleAssignments.Concat(roleAssignments);
+        }
+
+        // If a user has a regional role, remove it unless we're removing the ICB role in this update
+        if (hasRegionalPermission && !hasIcbScopedPermission)
```
pata9, line 217: I think this is the key clause missing a test?

src/api/Nhs.Appointments.Persistance/UserStore.cs
```diff
@@ -101,29 +104,7 @@ public async Task UpdateUserRegionPermissionsAsync(string userId, string scope,
...
-            updatedRoleAssignments = updatedRoleAssignments
-                .Where(ra => ra.Role != regionalUserRole)
-                .Concat(roleAssignments);
-        }
-        // No existing regional permission - add it
-        else
-        {
-            updatedRoleAssignments = updatedRoleAssignments.Concat(roleAssignments);
-        }
+        var updatedRoleAssignments = GetUpdatedPermissions(originalDocument.RoleAssignments, scope, roleAssignments, RegionalUserRole, IcbUserRole);
```
jsed-nhs, line 107: Does the calling order for Regional vs ICB in the args matter? They are called in a different order in other uses. Is PrimaryRole and ConflictingRole the best param names for this?

jsed-nhs (changes requested):

src/api/Nhs.Appointments.Persistance/UserStore.cs
ste-banks: It does matter yeah. For example if we were updating a Regional permission but passed it in as the second argument instead of the first, it would add a second regional permission to the user when they're only allowed one. Would also be the same for ICB scoped permissions if we passed them in the wrong way around. 

Yeah I'm not 100% happy with the naming myself if I'm being honest. PrimaryRole in this instance is the role we're adding, removing or updating so ICB or Region and the conflicting is the opposite of that. It would be conflicting if we're adding an ICB and the user already has a Region permission because it would need removing during the same update. More than happy to rename these to something else and open to suggestion what those could be.

pata9 (approved):

jsed-nhs (approved):

PR #943, opened 2025-08-27
APPT-XXX: Remove API Client Project

pata9:
# Description

API Client has been sat stale for a long time, removing to tidy up, we can bring this back if needed in the future. 

Fixes # (issue)

# Checklist:


sambiramairelogic (changes requested):
Sorry, I might be being a bit thick here but why are we deleting loads of integration tests? Shouldn't this only be deleting the Api.Client.UnitTests project?

pata9:
> > Sorry, I might be being a bit thick here but why are we deleting loads of integration tests? Shouldn't this only be deleting the Api.Client.UnitTests project?
> 
> That was not the intention :D Very good spot :D

Should be safe now!

PR #948, opened 2025-08-28
Appt 1178 - Cancel day page which is hooked up to the new endpoint

ste-banks:
# Description

This also includes the front end additions for APPT-1176

- Adding content to the cancel a day page
- Hooking it up to the new GetDaySummary and CancelDay endpoints
- Unit tests
- E2e tests will come later as further screens in the cancel day journey are yet to be developed

Fixes # (issue)

# Checklist:

### DIFF
--- a/src/client/src/app/site/[site]/cancel-day/cancel-day-form.test.tsx
+++ b/src/client/src/app/site/[site]/cancel-day/cancel-day-form.test.tsx
@@ -1,0 +1,93 @@
+import { render, screen, fireEvent } from '@testing-library/react';
+import CancelDayForm from './cancel-day-form';
+import { useRouter } from 'next/navigation';
+import { parseToUkDatetime } from '@services/timeService';
+
+jest.mock('next/navigation', () => ({
+  useRouter: jest.fn(),
+}));
+
+jest.mock('@services/timeService', () => ({
+  parseToUkDatetime: jest.fn(),
+}));
+
+jest.mock('@components/session-summary-table', () => ({
+  SessionSummaryTable: ({ tableCaption }: { tableCaption: string }) => (
+    <div data-testid="session-summary">{tableCaption}</div>
+  ),
+}));
+
+const mockReplace = jest.fn();
+
+const defaultProps = {
+  date: '2025-01-01',
+  siteId: 'site-123',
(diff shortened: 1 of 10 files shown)


src/client/src/app/site/[site]/cancel-day/cancel-day-form.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useRouter: jest.fn(),
+}));
+
+jest.mock('@services/timeService', () => ({
+  parseToUkDatetime: jest.fn(),
+}));
+
+jest.mock('@components/session-summary-table', () => ({
+  SessionSummaryTable: ({ tableCaption }: { tableCaption: string }) => (
+    <div data-testid="session-summary">{tableCaption}</div>
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/cancel-day/cancel-day-form.test.tsx:18]: Whilst it can sometimes be necessary, using `data-testid` is generally considered a code smell because if you can't find an element without using its id then that element is probably not accessible. 

In this case, it's totally redundant as the below assertion which uses it can just be rewritten as 
image
then the data-testid can be removed

src/client/src/app/site/[site]/cancel-day/cancel-day-form.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    expect(mockReplace).toHaveBeenCalledWith(
+      `/site/site-123/view-availability/week?date=2025-01-01`,
+    );
+  });
+
+  it('allows selecting Yes and shows confirmation step', async () => {
+    const { user } = render(<CancelDayForm {...defaultProps} />);
+
+    await user.click(screen.getByLabelText(/yes, i want to cancel/i));
+    await user.click(screen.getByRole('button', { name: /continue/i }));
```
Comment (sambiramairelogic) [inline src/client/src/app/site/[site]/cancel-day/cancel-day-form.test.tsx:84]: Bit of a minor one, but there's a performance hit to using regex so I'd try to only use it where necessary. We have quite a lot of tests now so probably worth noting. Most (all?) of the tests in this file still pass without needing the pattern matching - 
image

sambiramairelogic (changes requested):
Left a few suggestions, the only really blocking one is the feature toggle / permissions gate :)

sambiramairelogic (approved):

PR #966, opened 2025-09-04
APPT-1259 Make release pipeline INT dependency optional.

jsed-nhs:
# Description

Rename isHotfix -> ignoreIntDeployDependency and make it an optional param in the release pipeline alone. Defaulted to false. Can be overridden to be true if required in release pipeline for updated releases (due to cherry-picked hotfixes) that need deploying straight to Stag without Int (as Int has moved on since...)

Fixes # (issue)

# Checklist:


jsed-nhs:
This is not needed, can be done by skipping stage in Azure.

PR #975, opened 2025-09-09
Appt 1276 staging login issue with form action button

GLSM1-1:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

# Checklist:

### DIFF
--- a/src/client/src/app/lib/logging/logger.ts
+++ b/src/client/src/app/lib/logging/logger.ts
@@ -1,0 +1,25 @@
+type LogLevel = 'info' | 'warn' | 'error';
+
+export function logEvent(
+  level: LogLevel,
+  event: string,
+  payload: Record<string, any>,
+  context = 'App',
+) {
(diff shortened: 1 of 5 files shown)


src/client/src/app/lib/logging/logger.ts
```diff
@@ -0,0 +1,25 @@
+type LogLevel = 'info' | 'warn' | 'error';
+
+export function logEvent(
+  level: LogLevel,
+  event: string,
+  payload: Record<string, any>,
```
sambiramairelogic, line 6: You've probably also noticed this generates a warning because of the use of `any` which is a smell in typescript. Looks like in every use case of this func we're passing it an object, so we have two options here: 

- You could just change `any` to `unknown`, which tells Typescript we are explicitly declaring this as an unknown type.
- Or to do this a more proper Typescript way, we could label this func with a generic type, along the lines of: 
image

This should work with all the current places you're using this

PR #991, opened 2025-09-12
APPT-1350: Created availability edit proposal endpoint

Dave-O-NHS:
# Description

This PR adds a new function "ProposeAvailabilityChangeFunction" to evaluate the impact of session changes.

What’s New

- Function accepts the target session and requested change details.
- Runs the recalculation process to simulate the impact.
- Returns two counts:
  - Bookings supported – how many remain valid with the change.
  - Bookings unsupported – how many would be invalidated.

Why

This provides a safe way to preview the effect of a session update, helping avoid unintended disruptions to bookings.

# Checklist:


pata9:
> # Description
> Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
> 
> Fixes # (issue)
> 
> # Checklist:
> * [ ]  My work is behind a feature toggle (if appropriate)
> * [ ]  If my work is behind a feature toggle, I've added a full suite of tests for both the ON and OFF state
> * [ ]  The ticket number is in the Pull Request title, with format "APPT-XXX: My Title Here"
> * [ ]  I have ran npm tsc / lint (in the future these will be ran automatically)
> * [ ]  My code generates no new .NET warnings (in the future these will be treated as errors)
> * [ ]  If I've added a new Function, it is disabled in all but one of the terraform groups (e.g. http_functions)
> * [ ]  If I've added a new Function, it has both unit and integration tests. Any request body validators have unit tests also
> * [ ]  If I've made UI changes, I've added appropriate Playwright and Jest tests

Can you add a description of the change please

src/api/Nhs.Appointments.Api/Functions/AvailabilityChangeProposalFunction.cs
```diff
@@ -0,0 +1,60 @@
...
+            request.SessionMatcher,
+            request.SessionReplacement);
+
+        return ApiResult<AvailabilityChangeProposalResponse>.Success(
+            new AvailabilityChangeProposalResponse(
+                recalculations.ReallocatedBookingsCount, 
+                recalculations.UnaccommodatedBookingsCount
+            )
+        );
+    }
```
sambiramairelogic, line 59: lgtm so far, but I wonder if we should be returning a failure response if the `SessionMatcher` doesn't find a matching session?

src/api/Nhs.Appointments.Core/BookingAvailabilityStateService.cs
```diff
@@ -162,6 +173,39 @@ private async Task<BookingAvailabilityState> BuildState(string site, DateTime fr
...
+    {
+        var (bookings, sessions) =
+                await FetchData(site, from, to, returnType);
+
+        if (matcher is not null && replacement is not null)
+        {
+            var matchedSession = sessions.FirstOrDefault(s =>
+                s.From.TimeOfDay == matcher.From.ToTimeSpan() &&
+                s.Until.TimeOfDay == matcher.Until.ToTimeSpan() &&
+                matcher.Services.All(ms => s.Services.Contains(ms)));
```
sambiramairelogic, line 187: I think we might need to match on duration & capacity here as well?

src/api/Nhs.Appointments.Api/Functions/AvailabilityChangeProposalFunction.cs
Dave-O-NHS: I think that's a good idea, good spot

src/api/Nhs.Appointments.Core/BookingAvailabilityStateService.cs
Dave-O-NHS: We are now matching with duration and capacity as well. 
I had a look into that IEquatable, then I realized that we are not comparing same objects types, so stayed with manual matching.

ste-banks (changes requested):

PR #1006, opened 2025-09-16
APPT-1372/APPT-1376 - A couple of minor fixups for cancel a day

ste-banks:
# Description

- Only cancelling bookings on a day which don't have an already 'Cancelled Status'
- This should fix the numbers displayed on the front end as well
- Only returning the newly cancelled bookings with contact details to prevent duplicate cancelled notifications being sent
- Filtering out bookings with contact details on the front end when displaying bookings without them - this list now also includes bookings without contact details that were already cancelled before the cancel a day functionality was ran (assuming the reason was CancelledBySite)
- Including orphaned appointments in the summary booking count when cancelling a day
- Updating the appointments table to state 'Not provided' when no contact details are provided

Fixes # (issue)

# Checklist:

### DIFF
--- a/src/api/Nhs.Appointments.Persistance/BookingCosmosDocumentStore.cs
+++ b/src/api/Nhs.Appointments.Persistance/BookingCosmosDocumentStore.cs
@@ -322,19 +322,22 @@ public class BookingCosmosDocumentStore(
 
     public async Task<(int cancelledBookingsCount, int bookingsWithoutContactDetailsCount, List<Booking> bookingsWithContactDetails)> CancelAllBookingsInDay(string site, DateOnly date)
     {
+        var startOfDay = date.ToDateTime(TimeOnly.MinValue);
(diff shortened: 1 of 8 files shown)


src/api/Nhs.Appointments.Persistance/BookingCosmosDocumentStore.cs
```diff
@@ -328,11 +328,13 @@ public async Task DeleteBooking(string reference, string site) => await Task.Whe
             var endOfDay = date.AddDays(1).ToDateTime(TimeOnly.MinValue).AddTicks(-1);
 
             var bookings = await GetInDateRangeAsync(startOfDay, endOfDay, site);
+            // Filter out already cancelled bookings so we don't send duplicate notifications
+            var activeBookings = bookings.Where(b => b.Status != AppointmentStatus.Cancelled).ToList();
```
pata9, line 332: We could move this query to the database?

pata9 (approved):

PR #1029, opened 2025-09-23
APPT-1389 update to terraform scaling out rules for the web application

DACA16:
# Description

1. Update to the scale out settings for the MYA web app
- This is increasing from 1 running instance (actually 2 in production) to 4 minimum instances in production
- The default is set to 6 and the minimum is set to 4
- So:
   - Idle 6 instances will be available
   - Low load but above minimum - scales down to 4
   - High load it can scale up to 20

2. Scale out and in rules
- Adopting the recommended settings from our performance test findings

This is a temporary increase, so only adjusting for STAG and PROD

# Checklist:

- [ N/A ] My work is behind a feature toggle (if appropriate)
- [ N/A ] If my work is behind a feature toggle, I've added a full suite of tests for both the ON and OFF state
- [ Y ] The ticket number is in the Pull Request title, with format "APPT-XXX: My Title Here"
- [ N/A ] I have ran npm tsc / lint (in the future these will be ran automatically)
- [ N/A ] My code generates no new .NET warnings (in the future these will be treated as errors)
- [ N/A ] If I've added a new Function, it is disabled in all but one of the terraform groups (e.g. http_functions)
- [ N/A ] If I've added a new Function, it has both unit and integration tests. Any request body validators have unit tests also
- [ N/A ] If I've made UI changes, I've added appropriate Playwright and Jest tests

### DIFF
--- a/infrastructure/resources/web_appservice.tf
+++ b/infrastructure/resources/web_appservice.tf
@@ -82,7 +82,94 @@ resource "azurerm_monitor_autoscale_setting" "nbs_mya_web_app_service_autoscale_
       maximum = var.web_app_service_plan_max_worker_count
     }
 
-    # CPU auto scale rule
+    # Scale out rules for MYA
+
+    # Memory scaling
+    rule {
+      metric_trigger {
+        metric_name        = "MemoryPercentage"
+        metric_resource_id = azurerm_service_plan.nbs_mya_web_app_service_plan.id
+        time_grain         = "PT1M"
+        statistic          = "Average"
+        time_window        = "PT3M"
+        time_aggregation   = "Average"
+        operator           = "GreaterThan"
+        threshold          = 70
+        metric_namespace   = "microsoft.web/serverfarms"
+      }
+
+      scale_action {
+        direction = "Increase"
+        type      = "ChangeCount"
+        value     = var.web_app_service_plan_scale_out_worker_count
+        cooldown  = "PT10M"
+      }
+    }
+
+    rule {
+      metric_trigger {
+        metric_name        = "MemoryPercentage"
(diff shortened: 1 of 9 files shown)


infrastructure/resources/web_appservice.tf
```diff
@@ -82,123 +82,143 @@ resource "azurerm_monitor_autoscale_setting" "nbs_mya_web_app_service_autoscale_
...
         metric_name        = "MemoryPercentage"
         metric_resource_id = azurerm_service_plan.nbs_mya_web_app_service_plan.id
         time_grain         = "PT1M"
         statistic          = "Average"
-        time_window        = "PT5M"
+        time_window        = "PT10M"
         time_aggregation   = "Average"
         operator           = "GreaterThan"
-        threshold          = 50
+        threshold          = 40
```
pata9, line 139: This is a second rule for scaling out due to memory percentage?

pata9: I think your intention here is to scale down at 40?

DACA16: Looking at the advice from performance testing, it suggests scaling up by 2 and 3 instances:  **APPT-1389**

**Scale-Out Strategy (Aggressive Protection)**

**Memory-Based Scaling (3 tiers of protection):**

40% memory: +2 instances (moderate response)
55% memory: +3 instances (high response)
70% memory: +3 instances (critical response)

**CPU-Based Scaling (3 tiers of protection):**

20% CPU: +2 instances (early intervention)
40% CPU: +3 instances (medium load)
60% CPU: +3 instances (high load response)

pata9 (approved):

PR #1044, opened 2025-09-29
APPT-1299 - Adding booking batch size (when present) to reports

ste-banks:
# Description

Adding booking batch size (when present) to reports

Fixes # (issue)

# Checklist:


data/BookingsDataExtracts/BookingDataConverter.cs
```diff
@@ -42,6 +42,8 @@ public class BookingDataConverter(IEnumerable<SiteDocument> sites)
     public static string ExtractService(BookingDocument booking) => booking.Service;
 
     public static string ExtractCancelledDateTime(BookingDocument booking) => booking.Status == AppointmentStatus.Cancelled ? booking.StatusUpdated.ToString("yyyy-MM-ddTHH:mm:sszzz") : null;
+
+    public static int? ExtractBatchSize(BookingDocument booking) => booking.BookingBatchSize ?? null;
```
pata9, line 46: I think we should default to `1`

ste-banks: In the ticket AC is says to send it as null - also it was written by you ;)

PR #1059, opened 2025-10-03
APPT-1427 - add covid 5_11 and 12_17 to site by area function

pata9:
# Description

Release 2.6.1 enables NBS to query site by area for covid services in cohort 5_11 and 12_17

Fixes # (issue)

# Checklist:


sambiramairelogic:
Hate myself for even asking, but I guess this is technically a contract change because we're extending the list of permitted inputs. Do we have approved to do this without feature toggles/versioning? 
I assume yes, but have to ask

PR #1063, opened 2025-10-06
APPT-1441: manual notifier added to mya

pata9:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

# Checklist:

### DIFF
--- a/data/JobRunner/JobRunner.csproj
+++ b/data/JobRunner/JobRunner.csproj
@@ -1,0 +1,36 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net8.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <Content Update="appsettings.json">
+        <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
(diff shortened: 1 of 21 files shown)


data/JobRunner/JobRunner.csproj
```diff
@@ -0,0 +1,36 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net8.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
```
sambiramairelogic, line 7: Minor thing but the decision was taken awhile back that we wouldn't be using nullable reference types in MYA yet, so this should be disable atm. 
Personally I wanted to use them, but it'd be too big of a refactor so it's a future tech debt concern

data/JobRunner/JobRunner.csproj
```diff
@@ -0,0 +1,36 @@
...
+      <ProjectReference Include="..\..\src\api\Nhs.Appointments.Api\Nhs.Appointments.Api.csproj" />
+      <ProjectReference Include="..\..\src\api\Nhs.Appointments.Persistance\Nhs.Appointments.Persistance.csproj" />
+      <ProjectReference Include="..\DataExtract\DataExtract.csproj" />
+    </ItemGroup>
+
+  <ItemGroup>
+    <None Update="appsettings.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
sambiramairelogic, line 34: I might just be being thick but do we need both item groups for appsettings? The one above should suffice shouldn't it?

PR #1601, opened 2026-04-10
APPT-2222 CosmosOperationHelper handles ResponseMessage 429 in its statusCode way

jsed-nhs:
# Description

Handle ResponseMessage TooManyRequests response differently, as Cosmos does NOT throw a CosmosException in this case. Handle and throw custom exception so that it is picked up by the CosmosOperation retry logic correctly.

### DIFF
--- a/tests/Nhs.Appointments.Persistance.UnitTests/TypedDocumentCosmosStoreTests.cs
+++ b/tests/Nhs.Appointments.Persistance.UnitTests/TypedDocumentCosmosStoreTests.cs
@@ -457,8 +457,8 @@ public class TypedDocumentCosmosStoreTests
                 {
                     ContainerName = "test-container",
                     BackoffRetryType = BackoffRetryType.Linear,
-                    CutoffRetryMs = 10,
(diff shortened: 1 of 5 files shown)


tests/Nhs.Appointments.Persistance.UnitTests/TypedDocumentCosmosStoreTests.cs
```diff
@@ -1035,8 +1035,8 @@ public async Task
     [Theory]
     [InlineData(13, 35, 96, 261, 600)]
     [InlineData(10, 27, 73, 200, 500)]
-    [InlineData(7, 19, 51, 140, 350)]
-    [InlineData(5, 13, 36, 100, 250)]
+    // [InlineData(7, 19, 51, 140, 350)]
```
jsed-nhs, line 1105: Min is 10 ms, need to update these to higher values

pata9 (dismissed):

SimonHarropNHS (dismissed):

pata9 (approved):

PR #1606, opened 2026-04-13
APPT-2128 Update back links for change availability for daily, week and month and the reports page.

pchoi-al:
# Description

Updated the back links for the Before you continue page for each of the Day, Week, Month view as the issue was when clicking the back link redirects to an incorrect URL when on a external non MYA page or when the session times out. I have also fixed the back link for the Reports Page.

The previous implementation relied on the browser's history stack (router.back()), which was unreliable in two key scenarios:
- When a session expired and the user logged back in, the redirect cycle cleared the history stack, causing the "Back" link to break or redirect to an incorrect URL.
- If the user landed on the page from an external source or a non-MYA page, the "Back" button would attempt to navigate the user out of the application context.

I have replaced the browser-history-dependent navigation with a redirect pattern. By passing the originating path via the URL, the "Back" link now:
- Persists through sessions: It remains functional even after a logout/login cycle.
- It always returns the user to a valid page, with a fallback to the site List if no specific origin is detected.

Fixes # (issue)

# Checklist:

### DIFF
--- a/src/client/src/app/lib/components/nhs-page.test.tsx
+++ b/src/client/src/app/lib/components/nhs-page.test.tsx
@@ -1,4 +1,5 @@
-import { render, screen } from '@testing-library/react';
+import { render, screen, act } from '@testing-library/react';
+import { ComponentProps } from 'react';
 import NhsPage from './nhs-page';
 import {
   fetchUserProfile,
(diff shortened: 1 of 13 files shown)


Dave-O-NHS (dismissed):

ste-banks (commented):
Looks good, could be worth adding some tests to ensure it's working as expected.

Dave-O-NHS (approved):

ste-banks (approved):

PR #1611, opened 2026-04-14
APPT-2110: Replace site locking with site+day locking for existing usages

SimonHarropNHS:
# Description

For the areas of the system that already used site locking, refactored this to use the combination of site+day in an attempt to reduce unnecessary collisions.

Fixes # (issue)

# Checklist:

### DIFF
--- a/src/api/Nhs.Appointments.Core/Bookings/BookingDayRange.cs
+++ b/src/api/Nhs.Appointments.Core/Bookings/BookingDayRange.cs
@@ -1,0 +1,9 @@
+
+namespace Nhs.Appointments.Core.Bookings;
+
+public class BookingDayRange(DateOnly date)
+{
+    public DateTime Start { get; } = new(date, new TimeOnly(0, 0));
+
+    public DateTime End { get; } = new(date, new TimeOnly(23, 59));
+}
--- a/src/api/Nhs.Appointments.Core/Bookings/BookingDayRange.cs
+++ b/src/api/Nhs.Appointments.Core/Bookings/BookingDayRange.cs
@@ -3,7 +3,7 @@ namespace Nhs.Appointments.Core.Bookings;
 
 public class BookingDayRange(DateOnly date)
 {
-    public DateTime Start { get; } = new(date, new TimeOnly(0, 0));
+    public DateTime Start { get; } = new(date, new TimeOnly(0, 0), DateTimeKind.Unspecified);
 
-    public DateTime End { get; } = new(date, new TimeOnly(23, 59));
+    public DateTime End { get; } = new(date, new TimeOnly(23, 59), DateTimeKind.Unspecified);
 }
--- a/src/api/Nhs.Appointments.Core/Concurrency/ISiteLeaseManager.cs
+++ b/src/api/Nhs.Appointments.Core/Concurrency/ISiteLeaseManager.cs
@@ -1,6 +1,7 @@
-﻿namespace Nhs.Appointments.Core.Concurrency;
+
+namespace Nhs.Appointments.Core.Concurrency;
 
 public interface ISiteLeaseManager
 {
-    ISiteLeaseContext Acquire(string site);
+    ISiteLeaseContext Acquire(string site, DateOnly date);
 }
--- a/tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
+++ b/tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
@@ -37,6 +37,7 @@ namespace Nhs.Appointments.Core.UnitTests
         [Fact]
         public async Task MakeBooking_AcquiresLock_WhenBooking()
         {
+            var expectedSiteId = Guid.NewGuid().ToString();
             var expectedFrom = new DateOnly(2077, 1, 1);
             var expectedUntil = expectedFrom.AddDays(1);
             SessionInstance[] availability =
@@ -46,10 +47,9 @@ namespace Nhs.Appointments.Core.UnitTests
 
             var booking = new Booking
             {
-                Site = "TEST", Service = "TSERV", From = new DateTime(2077, 1, 1, 10, 0, 0, 0), Duration = 10
+                Site = expectedSiteId, Service = "TSERV", From = new DateTime(2077, 1, 1, 10, 0, 0, 0), Duration = 10
             };
 
-            _siteLeaseManager.Setup(x => x.Acquire(It.IsAny<string>())).Returns(new FakeLeaseContext());
             _referenceNumberProvider.Setup(x => x.GetReferenceNumber(It.IsAny<string>())).ReturnsAsync("TEST1");
 
             var bookingQueryService = new BookingQueryService(_bookingsDocumentStore.Object, TimeProvider.System);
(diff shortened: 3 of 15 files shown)


tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
```diff
@@ -94,7 +94,7 @@ public async Task MakeBooking_RaisesNotificationEvent_WhenBooking()
                 Status = AppointmentStatus.Booked
             };
 
-            _siteLeaseManager.Setup(x => x.Acquire(It.IsAny<string>())).Returns(new FakeLeaseContext());
+            _siteLeaseManager.Setup(x => x.Acquire(It.IsAny<string>(), It.IsAny<DateOnly>())).Returns(new FakeLeaseContext());
```
jsed-nhs: Can we mock this out with the expected string and Date, rather than IsAny?

tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
```diff
@@ -37,6 +37,7 @@ public BookingWriteServiceTests()
         [Fact]
```
jsed-nhs, line 37: Can we add some tests in to prove the concurrency problem we are solving? i.e bookings made for the same site but on different days, take two different locks and they aren't the same.

src/api/Nhs.Appointments.Core/Concurrency/ISiteLeaseManager.cs
```diff
@@ -1,6 +1,7 @@
-﻿namespace Nhs.Appointments.Core.Concurrency;
+
+namespace Nhs.Appointments.Core.Concurrency;
 
 public interface ISiteLeaseManager
```
jsed-nhs, line 4: Should we name this interface something else now that it doesn't lock down the whole site?

src/api/Nhs.Appointments.Core/Bookings/BookingDayRange.cs
```diff
@@ -0,0 +1,9 @@
+
+namespace Nhs.Appointments.Core.Bookings;
+
+public class BookingDayRange(DateOnly date)
+{
+    public DateTime Start { get; } = new(date, new TimeOnly(0, 0));
```
jsed-nhs: Sonarcloud suggestions that you likely didn't introduce, but now need fixing :D

tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
jsed-nhs, line 37: Can we also have some datetime DST boundary tests to prove the datetime object stays consistent with local UK time. i.e a Booking comes in at 1:30am on the day the clocks change, which day does it take the lock for?

jsed-nhs (changes requested):
Can we also add some new integration tests that are NOT awaited asynchronous (currently they are all awaited) - that cover a concurrent example. i.e. A test where two bookings are made side-by-side (async unawaited) for the same site for different days. They should both succeed and not block each other (no need to test for the non-blocking though...?)

src/api/Nhs.Appointments.Core/Concurrency/ISiteLeaseManager.cs
SimonHarropNHS: It feels ok to me because I think of it as "the lease manager associated with a site", rather than being a list of the degrees of freedom for that lease, but happy to change it if you can think of a more salient name.

tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
SimonHarropNHS: Yep, done.

src/api/Nhs.Appointments.Core/Bookings/BookingDayRange.cs
SimonHarropNHS: Done

tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
SimonHarropNHS: Done

SimonHarropNHS: Done

src/api/Nhs.Appointments.Core/Concurrency/ISiteLeaseManager.cs
jsed-nhs: I think thats okay, but I would then name the acquire method to be explicit maybe? i.e. acquire Day lock? because maybe we will add more types of site lease locks in future (i.e lock down whole site/...,). Not too bothered if you don't agree!

tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
```diff
@@ -35,14 +35,16 @@ public BookingWriteServiceTests()
         }
 
         [Fact]
-        public async Task MakeBooking_AcquiresLock_WhenBooking()
+        public async Task MakeBooking_AcquiresLockForTheSiteIdAndDayOfTheBooking_WhenBooking()
         {
             var expectedSiteId = Guid.NewGuid().ToString();
             var expectedFrom = new DateOnly(2077, 1, 1);
             var expectedUntil = expectedFrom.AddDays(1);
```
jsed-nhs: Minor point, I can't see 'expectedUntil' being used anywhere?

tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
```diff
@@ -35,41 +35,259 @@ public BookingWriteServiceTests()
...
-            task.IsCompleted.Should().BeTrue();
+            // Act.
+            await bookingService.MakeBooking(booking1);
+            await bookingService.MakeBooking(booking2);
+
+            // Assert.
+            _siteLeaseManager.Verify(slm => slm.Acquire(expectedSiteId, expectedFrom), Times.Exactly(2));
+            _siteLeaseManager.VerifyNoOtherCalls();
+        }
+
```
jsed-nhs: Sorry to be annoying, can we also add a test that proves the DateTimes adhere to Local Uk time day and not UTC. i.e. a test where the date is 01-06-2026 00:30 (this is 31-05-2026 23:30 UTC) and that it takes a lock on June 1st not May 31st.

SimonHarropNHS: Ah cool, yes.

SimonHarropNHS: Added - I've changed the first test to show that the lock takes place against local time.

tests/Nhs.Appointments.Core.UnitTests/BookingWriteServiceTests.cs
SimonHarropNHS: I've fixed that.

PR #1616, opened 2026-04-15
APPT-2276: Ensure expire time is resolved before time is advanced

pata9:
# Description

Some unit tests in the cache service were timing out due to the advance time being called before the 'job'/'operation' had started and resolved when it should expire. Ensuring that the expire time is resolved before the Advance is called resolves the timeout

Fixes # (issue)

# Checklist:

### DIFF
--- a/tests/Nhs.Appointments.Core.UnitTests/CacheServiceTests.cs
+++ b/tests/Nhs.Appointments.Core.UnitTests/CacheServiceTests.cs
@@ -57,8 +57,8 @@ public class CacheServiceTests
     [Fact]
     public void LazySlidingCacheValuesSet()
     {
-        var cacheExpiration = _timeProvider.GetUtcNow().Add(ExpensiveOperationTimespan);
-        var cacheValue = _sut.GetLazySlidingCacheValue(DefaultCacheKey, DefaultOptions(cacheExpiration));
+        var expensiveOperationExpiration = _timeProvider.GetUtcNow().Add(ExpensiveOperationTimespan);
+        var cacheValue = _sut.GetLazySlidingCacheValue(DefaultCacheKey, DefaultOptions(expensiveOperationExpiration));
         Assert.False(cacheValue.IsCompleted);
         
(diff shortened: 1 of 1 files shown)


tests/Nhs.Appointments.Core.UnitTests/CacheServiceTests.cs
```diff
@@ -244,15 +250,16 @@ public async Task Standard_Lazy_Sliding_Behaviour_NotTriggered_Below_Threshold()
     [Fact]
     public async Task MultipleCallsWithinTimeframe_DoNotTriggerMultipleExpensiveOperations_Expiration()
     {
-        var call1 = _sut.GetLazySlidingCacheValue(DefaultCacheKey, DefaultOptions);
-        _ = _sut.GetLazySlidingCacheValue(DefaultCacheKey, DefaultOptions);
-        _ = _sut.GetLazySlidingCacheValue(DefaultCacheKey, DefaultOptions);
+        var cacheExpirationRound1 = _timeProvider.GetUtcNow().Add(ExpensiveOperationTimespan);
```
jsed-nhs: Is this changing the logic of the tests? this 'advance' wasn't here before, so not sure why now?

jsed-nhs (changes requested):
I want to give this a good look over to make sure the added code to the actual tests isn't changing their nature/assertions. i.e. make sure these aren't false positives

tests/Nhs.Appointments.Core.UnitTests/CacheServiceTests.cs
jsed-nhs: Sorry misunderstood purpose, ignore.

jsed-nhs (commented):
Happy for others to approve. Looks okay to me.

jsed-nhs (commented):
.

jsed-nhs (dismissed):

jsed-nhs (approved):

PR #1618, opened 2026-04-16
APPT-2128 Back link redirects to incorrect mya url after session timeout

pchoi-al:
# Description

This fixes an issue where the back link would redirect to an incorrect or truncated URL following a session timeout.

When a session timed out, the middleware was capturing the current URL as a returnUrl but failing to correctly handle nested query parameters (specifically for the Day, Week, and Month views). Characters like & were being interpreted by the browser as part of the login redirect rather than part of the destination path, causing the return URL to be cut short.

Updated middleware to use encodeURIComponent() when constructing the redirect URL. his ensures that the entire original path—including all query parameters (e.g., ?date=...&page=...)

Fixes # (issue)

# Checklist:

### DIFF
--- a/src/client/src/middleware.ts
+++ b/src/client/src/middleware.ts
@@ -2,7 +2,9 @@ import { NextResponse } from 'next/server';
 import type { NextRequest } from 'next/server';
 
 export function middleware(request: NextRequest) {
-  const pathAndQuery = `${request.nextUrl.pathname}${request.nextUrl.search}`;
+  // Encode the search params (the ?... part) so that ampersands (&)
+  // are treated as text, not as separators for the login page parameters.
+  const pathAndQuery = `${request.nextUrl.pathname}${encodeURIComponent(request.nextUrl.search)}`;
 
   const shouldHandle = !request.nextUrl.pathname.includes('/api/');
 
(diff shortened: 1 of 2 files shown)


src/client/src/middleware.ts
```diff
@@ -2,7 +2,9 @@ import { NextResponse } from 'next/server';
 import type { NextRequest } from 'next/server';
 
 export function middleware(request: NextRequest) {
-  const pathAndQuery = `${request.nextUrl.pathname}${request.nextUrl.search}`;
+  // Encode the search params (the ?... part) so that ampersands (&)
+  // are treated as text, not as separators for the login page parameters.
+  const pathAndQuery = `${request.nextUrl.pathname}${encodeURIComponent(request.nextUrl.search)}`;
```
jsed-nhs, line 7: We will need to somehow test this with our akamai front door work... I remember such issues only presenting themselves on Staging/Prod. Need to think how we test this...

pchoi-al: That's a fair point. Since Akamai sits in front of the app in Staging/Prod, it might handle the encoded characters or the Referer header differently. I've used encodeURIComponent to make the string as 'safe' as possible

jsed-nhs: I'm not sure we can test this easily, happy to approve but think want to delay merge till after 2.15 cut taken

pchoi-al: No problem approve it after 2.15

